PR #11, opened 2022-08-21
Enhancements based on testing with BSC and Polygon

peterbroadhurst:
Resolves #9
Resolves #10 
Resolves #12 
Implements https://github.com/hyperledger/firefly-transaction-manager/issues/22

- For #9
    - Set catchup block gap / page size default to 2500
    - Fall back to catchup mode if we are about to call `eth_getFilterLogs` to far back from the chain head
- For #10
    - Enhance block listener to handle gaps in JSON/RPC stream, by holding an in-memory view of canonical chain
- For #12 
    - Treat a negative error code, with an empty `message` in a JSON/RPC response as an error
- For https://github.com/hyperledger/firefly-transaction-manager/issues/22
    - Add `catchup` flag to `getListenerHWM`

PR #15, opened 2022-08-23
introduce mapping for block not found errors

shorsher:
Encountered this during my testing with avalanche. 

Instead of returning empty block info for an unknown/unvalidated block, some evm blockchains will return an error. Now, the ethereum connector can keep track of these different errors and treat them as if `blockInfo` is `nil`

 in a chain with #11


peterbroadhurst (approved):
👍

PR #21, opened 2022-08-29
surface `ErrorNotFound` in get_block_info

shorsher:
Instead of swallowing `ErrorNotFound` for `blockRPCMethods`, surface it for consumers to decide how to handle the error. Also updated the evmconnect blocklistener implementation to not exit the loop on this error.

### DIFF
--- a/internal/ethereum/blocklistener.go
+++ b/internal/ethereum/blocklistener.go
@@ -292,12 +292,15 @@ func (bl *blockListener) rebuildCanonicalChain() *list.Element {
 	var notifyPos *list.Element
 	for {
 		var bi *blockInfoJSONRPC
+		var reason ffcapi.ErrorReason
 		err := bl.c.retry.Do(bl.ctx, "rebuild listener canonical chain", func(attempt int) (retry bool, err error) {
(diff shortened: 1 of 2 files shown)


internal/ethereum/blocklistener.go
```diff
@@ -292,12 +292,15 @@ func (bl *blockListener) rebuildCanonicalChain() *list.Element {
 	var notifyPos *list.Element
 	for {
 		var bi *blockInfoJSONRPC
+		var reason ffcapi.ErrorReason
 		err := bl.c.retry.Do(bl.ctx, "rebuild listener canonical chain", func(attempt int) (retry bool, err error) {
-			bi, err = bl.c.getBlockInfoByNumber(bl.ctx, nextBlockNumber, false, "")
+			bi, reason, err = bl.c.getBlockInfoByNumber(bl.ctx, nextBlockNumber, false, "")
 			return true, err
```
peterbroadhurst, line 298: does this need to be `return reason != ffcapi.ErrorReasonNotFound, err` to avoid us retrying on `ErrorReasonNotFound`?

shorsher: ah yep, thanks for catching that.

nguyer (approved):

PR #29, opened 2022-09-02
Lower default catchup page size

nguyer:
Signed-off-by: Nicko Guyer <nicko.guyer@kaleido.io>


peterbroadhurst (approved):
Noting the reason here I believe is testing with public networks, where the number of transactions in 2000 blocks could be so large as to timeout after 30secs even with a fast JSON/RPC endpoint

PR #37, opened 2022-09-29
Handle alternative response text for known transaction errors

matthew1001:
Signed-off-by: Matthew Whitehead <matthew.whitehead@kaleido.io>


matthew1001:
On further inspection I think his PR is incorrect and the alternative text was being rendered from a different part of FF

Issue #23, opened 2022-08-29
Intermittent test failure in `TestCatchupThenRejoinLeadGroup`


peterbroadhurst:
https://github.com/hyperledger/firefly-evmconnect/runs/8078322411?check_suite_focus=true
[logs_104.zip](https://github.com/hyperledger/firefly-evmconnect/files/9447612/logs_104.zip)

```
time="2022-08-29T19:31:03Z" level=debug msg="ChainHead=212121"
time="2022-08-29T19:31:03Z" level=debug msg="ChainHead=212121"
time="2022-08-29T19:31:03Z" level=debug msg="ChainHead=212121"
time="2022-08-29T19:31:03Z" level=debug msg="ChainHead=212121"
coverage: 78.5% of statements
panic: test timed out after 30s

goroutine 32 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1788 +0x8e
created by time.goFunc
	/opt/hostedtoolcache/go/1.17.13/x64/src/time/sleep.go:180 +0x31

goroutine 1 [chan receive]:
testing.(*T).Run(0xc00016bd40, {0x82b5bb, 0x46bbd3}, 0x851b10)
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1307 +0x375
testing.runTests.func1(0xc00016bd40)
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1598 +0x6e
testing.tRunner(0xc00016bd40, 0xc00029fce0)
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1259 +0x102
testing.runTests(0xc0000c9700, {0xb355a0, 0x70, 0x70}, {0x4808ad, 0x8225c2, 0xb3bb00})
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1596 +0x43f
testing.(*M).Run(0xc0000c9700)
	/opt/hostedtoolcache/go/1.17.13/x64/src/testing/testing.go:1504 +0x51d
main.main()
	_testmain.go:347 +0x1f5

goroutine 142 [sleep]:
time.Sleep(0xf4240)
	/opt/hostedtoolcache/go/1.17.13/x64/src/runtime/time.go:193 +0x12e
[...]

PR #39, opened 2022-10-07
Fix TestCatchupThenRejoinLeadGroup intermittent failure

peterbroadhurst:
Fixes https://github.com/hyperledger/firefly-evmconnect/issues/23

Explanation in extra log statement in test


nguyer (approved):
Thanks for fixing this!

PR #46, opened 2022-12-19
Set protocolId in all transaction receipts

matthew1001:
As required by FireFly core PR https://github.com/hyperledger/firefly/pull/1111 this PR brings evmconnect into line with the updated `TransactionReceiptResponse` struct that adds `protocolId` to transaction receipts.

Signed-off-by: Matthew Whitehead <matthew.whitehead@kaleido.io>


peterbroadhurst (approved):
Looks great - noting we need to pull in a release from FFTM before merge

PR #67, opened 2023-03-30
Wrap all "execution reverted" calls with error code FF23021

### LINKED ISSUE #65: Reverted calls/transactions should return an error message with a recognizable error code
Currently, if you execute an invalid query/transaction (such as calling a method that does not exist), evmconnect returns a very generic `execution reverted` message:

For ease of parsing, these messages should be wrapped with some more description and a recognizable error code. In ethconnect, for example, this same scenario produces an error that reads `FFEC100148: Call failed: execution reverted`.

awrichar:
Fixes #65


jimthematrix (approved):
👍

PR #68, opened 2023-04-03
Fix image manifests

nguyer:


awrichar:
So... my understanding is that `buildx` now uses a new manifest format that includes "provenance" info by default. Apparently when pulling an image, Docker is more strict about looking for a container that matches the current platform - so lately the amd64 evmconnect builds have been acceptable to amd64 systems but not to arm64, for example, with an error like `no matching manifest for linux/arm64/v8 in the manifest list entries`.

Disabling the "provenance" info during the push somehow alters the manifest and causes it to be OK to pull a non-matching image, so an ARM system can pull and use an amd64 image.

Correct me if anything is inaccurate here, but overall this LGTM for the time being.

awrichar (approved):

awrichar:
Some more context just to clarify the exact issues (currently) with trying to pull an evmconnect Docker on ARM hardware - you can pull if you manually specify the platform, but cannot pull otherwise.
```
andrew@MacBook-Pro-6:~/workspace/firefly$ docker pull ghcr.io/hyperledger/firefly-evmconnect:head
head: Pulling from hyperledger/firefly-evmconnect
no matching manifest for linux/arm64/v8 in the manifest list entries

andrew@MacBook-Pro-6:~/workspace/firefly$ docker pull --platform=amd64 ghcr.io/hyperledger/firefly-evmconnect:head
head: Pulling from hyperledger/firefly-evmconnect
3689b8de819b: Pull complete 
60675311ebf0: Pull complete 
60d22454596d: Pull complete 
0c01b9fcb88b: Pull complete 
Digest: sha256:131bca9bc420507edddb20b60e54d217c32ddbc90de0088c1d5111f2326fc907
Status: Downloaded newer image for ghcr.io/hyperledger/firefly-evmconnect:head
WARNING: image with reference ghcr.io/hyperledger/firefly-evmconnect was found but does not match the specified platform: wanted linux/amd64, actual: linux/arm64/v8
ghcr.io/hyperledger/firefly-evmconnect:head
```
Again, I presume this fix makes the first command work by making it less picky about which image to pull.

Issue #60, opened 2023-03-14
Return a better error when a transaction runs out of gas


awrichar:
If I submit a transaction with an insufficient `gas` value, and the EVM ends up running out of gas, it will simply stop executing and return a status of `0` in the transaction receipt. The transaction will still be included in a block, but will not be completed.

While the reason is not provided on the receipt in this case, it can be found by inspecting the transaction trace obtained by `debug.traceTransaction()`, if implemented: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracetransaction

The failed step of the transaction will include an `error` field, such as the `error: "out of gas"` entry seen in this trace: [trace.txt](https://github.com/hyperledger/firefly-evmconnect/files/10969700/trace.txt)

It seems that some tools such as etherscan are able to extract this information for the user: https://etherscan.io/tx/0xda8c0b80d8e240a83c8f6b067c4656babeb13e8e0ece4fd4292aa06252f1285c

It would be convenient if evmconnect could also provide this functionality.

hosie:
This would also help with other error conditions.  For example, if I have a `require` in my .sol such as
```solidity
require(balances[msg.sender] >= amount, "Not enough tokens");
```

and that causes the transaction to be reverted, then the error message is not returned on the receipt.  

Following @awrichar suggestion above, if I get the transaction hash from the receipt (e.g. using a local firefly stack)
```bash
TRANSACTION_HASH=$(curl -s "http://127.0.0.1:5000/api/v1/namespaces/default/operations/${OPERATION_ID}?fetchstatus=true" | jq -r .detail.transactionHash)
```

Then use that to read the transaction trace, convert the `returnValue` from hex to ascii (e.g. using `xxd`) and extract the strings
```bash
curl -s localhost:5100 -X POST --header 'Content-type: application/json' --data '{"jsonrpc":"2.0", "method":"debug_traceTransaction", "params":["'${TRANSACTION_HASH}'", {}], "id":1}' | jq  -r .result.returnValue | xxd -r -p | strings
```

then I do see the output
```
Not enough tokens
```

So this a) gives confidence in feasibility of the suggested fix and b) may serve as a short term workaround for anyone blocked by this in the meantime.

I have tried this on a local geth, provisioned by `ff init` and also on a hosted quorum node and it worked fine on both.

hosie:
[...]

PR #81, opened 2023-04-29
Error receipt

hosie:
Add `errorMessage` field to the receipt when a transaction fails by calling `debug_traceTransaction`.  Given that this is not certain to be implemented on all clients, we tolerate any errors and just omit the errorMessage from the receipt .

This has been tested on besu and quorum with a `require` statement in the contract causing the revert.  As yet I have been unable to prove that this is also effective in the case where there is not enough gas.  So not claiming that this fixes https://github.com/hyperledger/firefly-evmconnect/issues/60 just yet but it is a useful fix nonetheless.

### DIFF
--- a/internal/ethereum/get_receipt.go
+++ b/internal/ethereum/get_receipt.go
@@ -22,12 +22,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"unicode"
 
 	"github.com/hyperledger/firefly-common/pkg/fftypes"
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
 	"github.com/hyperledger/firefly-evmconnect/internal/msgs"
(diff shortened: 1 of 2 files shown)


internal/ethereum/get_receipt.go
```diff
@@ -97,6 +119,75 @@ func ProtocolIDForReceipt(blockNumber, transactionIndex *fftypes.FFBigInt) strin
...
+		]`))
+	if err != nil {
+		return nil, err
+	}
+
+	value, err := decodeCallDataHex(ctx, parsedABI.Functions()["Error"], returnValue)
+	if err != nil {
+		return nil, err
+	}
+	errorMessage := value.Children[0].Value
```
matthew1001, line 186: I assume we can never get here with `value.Children` having 0 elements because `decodeCallDataHex()` would return an error if the ABI didn't match?

matthew1001 (approved):
LGTM

PR #90, opened 2023-06-24
pick up rich query with groupby filter

Chengxuan:
Picks up https://github.com/kaleido-io/firefly-transaction-manager/pull/3


peterbroadhurst:
Going to pull in a separate commit of FFTM, as there are other updates

PR #112, opened 2024-02-19
Improve handling of custom errors in receipts - capture returnValue

peterbroadhurst:
See https://github.com/hyperledger/firefly/issues/1466 for details 

This PR is the first step of capturing the returnValue in cases it cannot be parsed as a default error


internal/ethereum/get_receipt_test.go
```diff
@@ -174,7 +174,7 @@ const sampleTransactionTraceBesu = `{
 				"0000000000000000000000000000000000000000000000000000000000000000"
 			],
 			"storage": {},
-			"reason": "0x8c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000114e6f7420656e6f75676820746f6b656e73000000000000000000000000000000"
```
peterbroadhurst, line 177: This was not a correct error - there is a missing `0` in the default `Error(string)` function selector `0x08c379a0`

Chengxuan (approved):
LGTM

internal/ethereum/get_receipt_test.go
matthew1001: @peterbroadhurst some nodes return the error in that format (i.e. with an odd-number of hex digits) and assume that you will pad with a single `0` client-side if it's odd. The new value you've used is valid, but the reason for include this particular reason was to emulate nodes which do that.

PR #118, opened 2024-03-07
feature: Support for Hedera EVM block identifiers

SamMayWork:
ref: https://discord.com/channels/905194001349627914/928377875827154984/1219628950393847860 (thread on Hyperledger discord on Hedera support with EVM Connect)

TL;DR: Hedera is a hashgraph not a blockchain (though is EVM compatible), and uses non-standard 'block' hashes which are 48 bytes rather than 32 bytes, EVM Connect cannot handle this currently. To promote integration with other ecosystems, Hedera expects that these 48 byte block headers be truncated to 32 bytes, this is a common feature in the Hedera ecosystem. 

https://hips.hedera.com/hip/hip-415 - Hedera introduction of blocks (mentioning 48 bytes being referenced by the 32 byte prefix)

https://hips.hedera.com/hip/hip-482 - JSON RPC specification for Hedera

> **Block Hashes**
> 
> Ethereum uses 32-byte hashes for blocks while Hedera uses 48-byte hashes. To compute a 32-byte Ethereum hash, take the leading 32 bytes from a 48-byte Hedera hash.


SamMayWork:
Once pre-req PR is merged we should be able to remove the custom go mod overrides.

go.mod
```diff
@@ -99,3 +99,5 @@ require (
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hyperledger/firefly-signer => github.com/SamMayWork/firefly-signer v0.0.0-20240307121035-e654fb9ef55b
```
SamMayWork, line 103: Will be removed after the pre-req changes are merged

peterbroadhurst: Please see review comments on https://github.com/hyperledger/firefly-signer/pull/61

internal/ethereum/blocklistener.go
```diff
@@ -140,6 +144,28 @@ func (bl *blockListener) listenLoop() {
...
+					continue
+				}
+
+				if bl.nonStandardBlockHashSizeResolutionMethod == "" {
+					log.L(bl.ctx).Warnf("No non-standard hash length resolution method was provided, defaulting to truncation")
+					bl.nonStandardBlockHashSizeResolutionMethod = "truncate"
+				}
+
+				if bl.nonStandardBlockHashSizeResolutionMethod == "truncate" {
+					h, err = h.Truncate(32)
```
peterbroadhurst, line 160:
```suggestion
h = ethtypes.HexBytes0xPrefix(h[0:32])
```
... combined with above suggestion

internal/msgs/en_config_descriptions.go
```diff
@@ -27,17 +27,19 @@ var ffc = func(key, translation string, fieldType string) i18n.ConfigMessageKey
...
+	ConfigEventsFilterPollingInterval              = ffc("config.connector.events.filterPollingInterval", "The interval between polling calls to a filter, when checking for newly arrived events", i18n.TimeDurationType)
+	ConfigTxCacheSize                              = ffc("config.connector.txCacheSize", "Maximum of transactions to hold in the transaction info cache", i18n.IntType)
+	ConfigMaxConcurrentRequests                    = ffc("config.connector.maxConcurrentRequests", "Maximum of concurrent requests to be submitted to the blockchain", i18n.IntType)
+	ConfigAllowNonStandardBlockHashSize            = ffc("config.connector.allowNonStandardBlockHashSize", "Whether to permit block hases of a non-standard (i.e. not 32 bytes) length", i18n.BooleanType)
+	ConfigNonStandardBlockHashSizeResolutionMethod = ffc("config.connector.nonStandardBlockHashSizeResolutionMethod", "How to resolve block hashes that are non-standard (i.e. not 32 bytes) in length. Currently only 'truncate' is supported.", i18n.StringType)
```
peterbroadhurst, line 44: I'm not sure I understand yet the intention here, and that suggests to me users of this setting might struggle to.

Is the intention here `blockHashByteLength`?

e.g. you are instructing the whole code stack that all block hashes should be X bytes, where the default is 32?

If so, what code is affected that needs this? And why? 

Will try and answer from reading through the rest of the code

internal/ethereum/blocklistener.go
```diff
@@ -140,6 +144,28 @@ func (bl *blockListener) listenLoop() {
 		update := &ffcapi.BlockHashEvent{GapPotential: gapPotential}
 		var notifyPos *list.Element
 		for _, h := range blockHashes {
+			if len(h) != 32 {
```
peterbroadhurst, line 147: Sorry @SamMayWork I couldn't understand in the end the problem and the solution. Could do with some help.

32 is the length of a "standard" Ethereum block hash. So if we get a standard one, it will be 32 bytes in length.

You are working with a EVM implementation against a ledger that uses a 48 byte block hash - so in that case, the length here would be 48.

Why would we truncate that? Surely we just need to acommodate that it's 48 bytes, and there will be some number of places in the code that need to be relaxed to avoid the assumption of 32 bytes. Maybe for example there are DB migration tables etc.

peterbroadhurst (changes requested):
Need a little more info on the solution you're proposing, and the alternatives you considered @SamMayWork if that's ok

internal/ethereum/blocklistener.go
SamMayWork: > Why would we truncate that? Surely we just need to acommodate that it's 48 bytes, and there will be some number of places in the code that need to be relaxed to avoid the assumption of 32 bytes. Maybe for example there are DB migration tables etc.

There was a brief [Discord discussion](https://discord.com/channels/905194001349627914/928377875827154984/1219628950393847860) on this, although in retrospect I don't think we considered what other alternative approaches we could use here.

The reason for using truncation here instead of changing the code everywhere to permit the change in header size was purely to reduce the scope of the changes required. I think the view was that Hedera _is_ EVM compatible but also a little bit weird, and it might make more sense to truncate the hashes so there's no need to change existing code for a single use-case.

We could use this PR to permit non-standard block header sizes everywhere if that's something strategically we think EVM Connect should be able to do.

internal/ethereum/blocklistener.go
SamMayWork: Yep agreed here - if we go with this approach, we'll do this and then remove the requirement on the firefly-signer PR.

peterbroadhurst:
> fix: Allow for non-standard block header lengths

Maybe worth changing to:

"feature: Support for Hedera EVM block identifiers"

internal/ethereum/blocklistener.go
peterbroadhurst:
```suggestion
			if len(h) == 48 {
```

internal/ethereum/blocklistener.go
```diff
@@ -140,6 +142,16 @@ func (bl *blockListener) listenLoop() {
...
+				if !bl.hederaCompatibilityMode {
+					log.L(bl.ctx).Errorf("Attempted to index block header with non-standard length: %d", len(h))
+					failCount++
+					continue
+				}
+
+				h = h[0:32]
```
peterbroadhurst, line 152: Note above to make this safe we need to be confident we have at least 32b

peterbroadhurst (approved):
Minor suggestion to tweak the length check before merge, and a request to change the PR title for the release when that happens.

EnriqueL8 (approved):
Approving and merging based on above approvals!

PR #128, opened 2024-04-08
If the revert reason is present in a receipt, skip the call to `debug_traceTransaction`

matthew1001:
Also re-instate odd-length hex padding with leading 0 (see https://github.com/hyperledger/firefly-evmconnect/pull/112)


Chengxuan (changes requested):
LGTM,  a few minor comments to avoid the amount of debug level logs being overwhelming

Chengxuan (approved):
LGTM

PR #139, opened 2024-06-10
picking up RPS throttle configuration for the connector

Chengxuan:
Picks up https://github.com/hyperledger/firefly-common/pull/141 so overall traffic rates to JSON-RPC endpoint can be throttled


matthew1001 (approved):
LGTM

PR #142, opened 2024-06-12
Faster block detection using WebSocket notifications

peterbroadhurst:
Add an option to use a WebSocket JSON/RPC connection _only for the block listener_, and as well as using `eth_newBlockFilter` with polling, to do an `eth_subscribe` to get shoulder taps as soon as new heads are available.

This reduces latency of detection of mining of transactions.


Chengxuan (approved):
LGTM

PR #147, opened 2024-08-19
use websocket backend for all rpc calls when it's turned on

Chengxuan:
When Websocket is turned on, the code currently only use it for receiving new heads. This PR updated the code to use Websockets for all RPC calls as well so a separate HTTP endpoint is not necessary.


matthew1001 (approved):
LGTM

PR #148, opened 2024-08-22
Decode numbers using json.Decoder

matthew1001:
If the `params` passed in for a contract deploy include very large numbers, `json.Unmarshal` to an `interface{}` doesn't work and results in an incorrect number being passed in to the contract's constructor.

To recreate the issue, if you checkout just the UT changes from this branch and run them it will fail.

This PR switches `json.Unmarshal` to `json.Decode`.

### DIFF
--- a/internal/ethereum/deploy_contract_prepare.go
+++ b/internal/ethereum/deploy_contract_prepare.go
@@ -17,6 +17,7 @@
 package ethereum
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/hex"
@@ -95,7 +96,9 @@ func (c *ethConnector) prepareDeployData(ctx context.Context, req *ffcapi.Contra
 	ethParams := make([]interface{}, len(req.Params))
 	for i, p := range req.Params {
 		if p != nil {
-			err := json.Unmarshal([]byte(*p), &ethParams[i])
+			decodedParam := json.NewDecoder(bytes.NewReader([]byte(*p)))
+			decodedParam.UseNumber()
+			err := decodedParam.Decode(&ethParams[i])
 			if err != nil {
 				return nil, nil, i18n.NewError(ctx, msgs.MsgUnmarshalParamFail, i, err)
 			}
--- a/internal/ethereum/deploy_contract_prepare_test.go
+++ b/internal/ethereum/deploy_contract_prepare_test.go
@@ -47,17 +47,12 @@ const samplePrepareDeployTX = `{
 				"internalType":" uint256",
 				"name": "y",
 				"type": "uint256"
-			},
-		    {
-			    "internalType":" string",
-			    "name": "x",
-			    "type": "string"
-		    }
+			}
 		],
 		"outputs":[],
 		"type":"constructor"
 	}],
-	"params": [ 4276993775, "some-text" ]
+	"params": [ 4276993775 ]
 }`
 
 const samplePrepareDeployTXLargeInputParams = `{
@@ -104,10 +99,8 @@ func TestDeployContractPrepareOkNoEstimate(t *testing.T) {
 	assert.Empty(t, reason)
 	assert.Equal(t, int64(1000000), res.Gas.Int64())
 
-	// Basic check that our input param 10000000000000000000000001 is in the TX data
+	// Basic check that our input param 4276993775 is in the TX data
 	assert.True(t, strings.Contains(res.TransactionData, "feedbeef"))
-	// Basic check that our input param "some-text" is in the TX data
-	assert.True(t, strings.Contains(res.TransactionData, "736f6d652d74657874"))
 }
 
 func TestDeployContractPrepareOkLargeInputParam(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/ethereum/deploy_contract_prepare.go
```diff
@@ -95,7 +96,9 @@ func (c *ethConnector) prepareDeployData(ctx context.Context, req *ffcapi.Contra
 	ethParams := make([]interface{}, len(req.Params))
 	for i, p := range req.Params {
 		if p != nil {
-			err := json.Unmarshal([]byte(*p), &ethParams[i])
+			decodedParam := json.NewDecoder(bytes.NewReader([]byte(*p)))
+			decodedParam.UseNumber()
```
EnriqueL8, line 100: From what I can read this stores the number in the string representation, have you checked that FireFly Signer supports this?

EnriqueL8: Just thinking of how the RLP encode code there would treat this

EnriqueL8 (commented):
Looks good - just one comment on testing with FireFly Signer and understanding if not only the number is there in RLP but also the things around it like where the number starts

internal/ethereum/deploy_contract_prepare.go
EnriqueL8: So it feels like it was just going wrong before it hit FireFly Signer https://github.com/hyperledger/firefly-signer/blob/5256116539ed83a23a52103f6260f4887abe0e85/pkg/abi/abiencode.go#L181

Chengxuan: I suspect we need to fix the issue in https://github.com/hyperledger/firefly-common/blob/main/pkg/fftypes/jsonany.go#L51

internal/ethereum/deploy_contract_prepare.go
```diff
@@ -95,7 +96,9 @@ func (c *ethConnector) prepareDeployData(ctx context.Context, req *ffcapi.Contra
...
+			err := decodedParam.Decode(&ethParams[i])
```
Chengxuan, line 101: I think we need to fix the issue in https://github.com/hyperledger/firefly-common/blob/main/pkg/fftypes/jsonany.go#L51 if the issue is how data gets unmarshalled

matthew1001: Agreed - and it probably needs a quick search of the code for other places where `json.Unmarshal()` is used as well

matthew1001: Going to prove out this fix with an end-to-end contract deploy first to check that the fix is good (beyond the UTs)

internal/ethereum/deploy_contract_prepare.go
peterbroadhurst: Trying to understand this. The code before seemed correct:
```go
// Parse the params into the standard semantics of Go JSON unmarshalling, with []interface{}
```

peterbroadhurst: The intent of this code should be to take:
- take raw `json.RawMessage` etc still input from `req.Params`
- for each, do a standard `json.Unmarshal()` to `interface{}` so Golang picks the type
- pass that down to `decodedParam.Decode` to do type-specific processing on

internal/ethereum/deploy_contract_prepare.go
peterbroadhurst: Sorry @matthew1001 - I'm not sure I understand why there's a concern with `json.Unmarshal()`

internal/ethereum/deploy_contract_prepare_test.go
```diff
@@ -54,6 +55,37 @@ const samplePrepareDeployTX = `{
...
+		    {
+			    "internalType":" string",
+			    "name": "x",
+			    "type": "string"
+		    }
+		],
+		"outputs":[],
+		"type":"constructor"
+	}],
+	"params": [ 10000000000000000000000001, "some-text" ]
```
peterbroadhurst, line 86: This input is broken in JSON, it's attempting to use a number that does not fit into int64

peterbroadhurst (changes requested):
...

internal/ethereum/deploy_contract_prepare.go
peterbroadhurst: Ok - I do think I now understand, but the right fix (which I think would probably include `json.Unmarshal`) would take time to investigate

peterbroadhurst (commented):
ok - sorry, after some misunderstanding I now understand what this PR is trying to achieve.

This PR is an enhancement to support:
- JSON input in the `params` array is able to contain JSON `number` elements larger than fit into an `int64`.

One part of the fix, is that `Golang` provides an option on `json.Decoder` (not available on `json.Unmarshal`) to decode `number`s into `json.Number` types (which are `strings`).

If that is done in this one piece of the stack, then it has implications for other places that need to be reviewed:
- `firefly`
- `firefly-signer`

With this done, a user choosing to generate a JSON input containing JSON `number` > that is possible to generate in Javascript/Java/Go serialization of numbers could be supported by the stack.

matthew1001:
Thanks for the comments @peterbroadhurst, @Chengxuan, @EnriqueL8 - I might grab some in-person time with @EnriqueL8 to discuss options & other places in the stack that will be affected. I'll circle back round on this PR once I've done that.

PR #153, opened 2024-08-27
Exploit new support in `firefly-signer` and `firefly-common` for large & scientific JSON numbers

matthew1001:
This PR does 2 things:

1. Pulls in the latest version of:
   - `firefly-common` (see especially PR https://github.com/hyperledger/firefly-common/pull/147)
   - `firefly-signer` (see especially PR https://github.com/hyperledger/firefly-signer/pull/76)
   - `firefly-transaction-manager` (see especially PR https://github.com/hyperledger/firefly-transaction-manager/pull/128)
3. Adds unit tests that exercise the new function introduced by those PRs

### DIFF
--- a/internal/ethereum/deploy_contract_prepare.go
+++ b/internal/ethereum/deploy_contract_prepare.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"strings"
 
-	"github.com/hyperledger/firefly-common/pkg/fftypes"
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
 	"github.com/hyperledger/firefly-evmconnect/internal/msgs"
@@ -93,10 +92,10 @@ func (c *ethConnector) prepareDeployData(ctx context.Context, req *ffcapi.Contra
 	}
 
 	// Parse the params into the standard semantics of Go JSON unmarshalling, with []interface{}
-	ethParams := make([]fftypes.JSONAny, len(req.Params))
+	ethParams := make([]interface{}, len(req.Params))
 	for i, p := range req.Params {
 		if p != nil {
-			err := json.Unmarshal([]byte(*p), &ethParams[i])
+			err := p.Unmarshal(ctx, &ethParams[i])
 			if err != nil {
 				return nil, nil, i18n.NewError(ctx, msgs.MsgUnmarshalParamFail, i, err)
 			}
--- a/internal/ethereum/deploy_contract_prepare.go
+++ b/internal/ethereum/deploy_contract_prepare.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"strings"
 
+	"github.com/hyperledger/firefly-common/pkg/fftypes"
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
 	"github.com/hyperledger/firefly-evmconnect/internal/msgs"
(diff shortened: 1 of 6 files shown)


internal/ethereum/deploy_contract_prepare.go
```diff
@@ -95,7 +95,7 @@ func (c *ethConnector) prepareDeployData(ctx context.Context, req *ffcapi.Contra
 	ethParams := make([]interface{}, len(req.Params))
 	for i, p := range req.Params {
 		if p != nil {
-			err := json.Unmarshal([]byte(*p), &ethParams[i])
+			err := p.Unmarshal(ctx, &ethParams[i])
```
EnriqueL8, line 98: For context: The type of `p` is `*fftypes.JSONAny` which has implemented an Unmarshal method

internal/ethereum/deploy_contract_prepare_test.go
```diff
@@ -39,19 +40,91 @@ const samplePrepareDeployTX = `{
 	"gas": 1000000,
 	"nonce": "111",
 	"value": "12345678901234567890123456789",
-	"contract": "0xfeedbeef",
+	"contract": "0xdeadbeef",
```
EnriqueL8, line 43: 🤣

internal/ethereum/deploy_contract_prepare_test.go
```diff
@@ -39,19 +40,91 @@ const samplePrepareDeployTX = `{
...
+		    {
+			    "internalType":" address",
+			    "name": "y",
+			    "type": "address"
+		    },
+		    {
+			    "internalType":" string",
+			    "name": "z",
+			    "type": "string"
+		    }
```
EnriqueL8, line 60: the hole nesting looks weird, probably just GitHub

EnriqueL8 (approved):
Looks good to me

EnriqueL8 (changes requested):
Wait don't we have to fix the invoke contract as well?

EnriqueL8:
You need the same fix in https://github.com/hyperledger/firefly-evmconnect/blob/6aff1e62adfcf4532cd3ccc501f2a7af4c75ea23/internal/ethereum/prepare_transaction.go#L111

internal/ethereum/deploy_contract_prepare_test.go
matthew1001: Yeah a slightly cheeky way to ensure the asserts in the tests don't match on the contract bytes when checking for integers that match `0xdeadbeef`

internal/ethereum/deploy_contract_prepare_test.go
matthew1001: yeah I think that's just github

matthew1001:
> You need the same fix in
> 
> https://github.com/hyperledger/firefly-evmconnect/blob/6aff1e62adfcf4532cd3ccc501f2a7af4c75ea23/internal/ethereum/prepare_transaction.go#L111

Good spot, will address that now

matthew1001:
> Could it also affect the gas price
> 
> https://github.com/hyperledger/firefly-evmconnect/blob/6aff1e62adfcf4532cd3ccc501f2a7af4c75ea23/internal/ethereum/send_transaction.go#L99
> 
> ?

I think this case is OK as we're specifically unmarshaling to a `HexInteger` not an `interface{}`

matthew1001:
@EnriqueL8 new commits added to handle the `prepare` path, with unit tests

EnriqueL8 (approved):
Looks good to me

EnriqueL8:
Re-Running Jobs to upload coverage

PR #154, opened 2024-09-05
Handle duplicate block number in the same getFilterChanges call

Chengxuan:
When the block hashes in the `eth_getFilterChanges` call has block number conflict, there is a edge case we don't handle.

This PR fixed the issue so that if the beginning block number has duplicate hashes in the same `eth_getFilterChanges` call, the logic should still be able to handle it.


peterbroadhurst (approved):
This is a great find @Chengxuan !

those single character changes are always the hardest ones to diagnose

PR #157, opened 2024-11-13
adding warning logs for rebuiding in memory chain

Chengxuan:
adding more logs to highlight the situation in which the in memory canonical chain is getting rebuilt

### DIFF
--- a/internal/ethereum/blocklistener.go
+++ b/internal/ethereum/blocklistener.go
@@ -381,14 +381,13 @@ func (bl *blockListener) handleNewBlock(mbi *minimalBlockInfo, addAfter *list.El
 // a recent block advertisement. So we need to work backwards to the last point of consistency with the current
 // chain and re-query the chain state from there.
 func (bl *blockListener) rebuildCanonicalChain() *list.Element {
-	log.L(bl.ctx).Warn("Rebuilding in-memory canonical chain")
-
 	// If none of our blocks were valid, start from the first block number we've notified about previously
 	lastValidBlock := bl.trimToLastValidBlock()
 	var nextBlockNumber int64
 	var expectedParentHash string
 	if lastValidBlock != nil {
 		nextBlockNumber = lastValidBlock.number + 1
+		log.L(bl.ctx).Warnf("rebuildCanonicalChain partial rebuild from block %d", nextBlockNumber)
 		expectedParentHash = lastValidBlock.hash
 	} else {
 		firstBlock := bl.canonicalChain.Front()
@@ -396,6 +395,7 @@ func (bl *blockListener) rebuildCanonicalChain() *list.Element {
 			return nil
 		}
 		nextBlockNumber = firstBlock.Value.(*minimalBlockInfo).number
+		log.L(bl.ctx).Warnf("rebuildCanonicalChain brand new in-memory canonical chain from block %d", nextBlockNumber)
 		// Clear out the whole chain
 		bl.canonicalChain = bl.canonicalChain.Init()
 	}
@@ -413,7 +413,7 @@ func (bl *blockListener) rebuildCanonicalChain() *list.Element {
 			}
 		}
 		if bi == nil {
-			log.L(bl.ctx).Debugf("Block listener canonical chain view rebuilt to head at block %d", nextBlockNumber-1)
+			log.L(bl.ctx).Infof("rebuildCanonicalChain successfully rebuilt the chain to the head block %d", nextBlockNumber-1)
 			break
 		}
 		mbi := &minimalBlockInfo{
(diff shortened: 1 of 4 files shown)


internal/ethereum/blocklistener.go
```diff
@@ -481,7 +484,10 @@ func (bl *blockListener) trimToLastValidBlock() (lastValidBlock *minimalBlockInf
 			break
 		}
 		lastElem = lastElem.Prev()
+	}
 
+	if startingNumber != nil && lastValidBlock != nil && *startingNumber != lastValidBlock.number {
+		log.L(bl.ctx).Warnf("Trimmed canonical chain to from block %d to block %d (total number of in memory blocks: %d)", startingNumber, lastValidBlock.number, bl.unstableHeadLength)
```
Chengxuan, line 490:
```suggestion
		log.L(bl.ctx).Warnf("Trimming canonical chain to from block %d to block %d (total number of in memory blocks: %d)", startingNumber, lastValidBlock.number, bl.unstableHeadLength)
```

internal/ethereum/event_enricher.go
```diff
@@ -84,6 +86,10 @@ func (ee *eventEnricher) filterEnrichEthLog(ctx context.Context, f *eventFilter,
 		}
 	}
 
+	if blockNumber < 0 || transactionIndex < 0 || logIndex < 0 {
```
Chengxuan, line 89: overflow check as spotted by the linter

internal/ethereum/blocklistener.go
```diff
@@ -426,7 +425,7 @@ func (bl *blockListener) rebuildCanonicalChain() *list.Element {
 		// It's possible the chain will change while we're doing this, and we fall back to the next block notification
 		// to sort that out.
 		if expectedParentHash != "" && mbi.parentHash != expectedParentHash {
-			log.L(bl.ctx).Debugf("Block listener canonical chain view rebuilt up to new re-org at block %d", nextBlockNumber)
+			log.L(bl.ctx).Warnf("rebuildCanonicalChain rebuild stopped at block: %d due to mismatch hash for parent block (%d): %s (expected: %s)", nextBlockNumber, nextBlockNumber-1, mbi.parentHash, expectedParentHash)
```
peterbroadhurst, line 428: Is this the change @Chengxuan ?

I see a move from regular logs otherwise, to including some function call semantics in the log names.
That's not a general approach this repo uses. There's pros and cons to it... but the main thing I'm concerned about is consistency.

Would you mind just making the logging change?

internal/ethereum/blocklistener.go
```diff
@@ -381,22 +381,21 @@ func (bl *blockListener) handleNewBlock(mbi *minimalBlockInfo, addAfter *list.El
...
 		nextBlockNumber = lastValidBlock.number + 1
+		log.L(bl.ctx).Warnf("rebuildCanonicalChain partial rebuild from block %d", nextBlockNumber)
 		expectedParentHash = lastValidBlock.hash
 	} else {
 		firstBlock := bl.canonicalChain.Front()
 		if firstBlock == nil || firstBlock.Value == nil {
 			return nil
 		}
 		nextBlockNumber = firstBlock.Value.(*minimalBlockInfo).number
+		log.L(bl.ctx).Warnf("rebuildCanonicalChain brand new in-memory canonical chain from block %d", nextBlockNumber)
```
peterbroadhurst, line 398: These should not be warn levels as far as I know. This is normal operation as far as I'm aware

peterbroadhurst (changes requested):
couple of small requests before this merges @Chengxuan

internal/ethereum/blocklistener.go
Chengxuan: I reverted other warn level logs to be info and left this one as `warn`. I think this one should be at `warn` level because we are throwing away the in memory chain here.

internal/ethereum/blocklistener.go
Chengxuan: I switched them back for consistency. I personally found it's easier to have a unique ID to filter out logs. 

From ops point of view, I need a string that have a good summary of what a function does when searching for logs, and I assume that's an important function of method names. Therefore, I prefer including function name in the collected logs. (no matter it's in log context or the message itself) other than inventing another way of summarize what the logs line is for.

Chengxuan:
Thanks @peterbroadhurst both comments addressed

internal/ethereum/blocklistener.go
peterbroadhurst: We discussed offline that this is a good thing to work out separately, and needs to consider that the logging framework does provide this information (if you switch to JSON logging you'll see it)

peterbroadhurst (approved):

PR #158, opened 2024-12-05
Restructure block listener retry configuration

Chengxuan:
Move the block listener retry configuration outside the `retry` attribute as that was for resty client retry. Having them together is confusing.


internal/ethereum/event_enricher.go
```diff
@@ -96,9 +96,9 @@ func (ee *eventEnricher) filterEnrichEthLog(ctx context.Context, f *eventFilter,
 			Signature:        signature,
 			BlockHash:        ethLog.BlockHash.String(),
 			TransactionHash:  ethLog.TransactionHash.String(),
-			BlockNumber:      fftypes.FFuint64(blockNumber),
-			TransactionIndex: fftypes.FFuint64(transactionIndex),
-			LogIndex:         fftypes.FFuint64(logIndex),
```
EnriqueL8, line 101: This is breaking change - with deprecated fields we keep it around for one release at least so we support both. Especially with a minor release.

EnriqueL8 (changes requested):

internal/ethereum/event_enricher.go
Chengxuan: this line here is for a linter fix.

Chengxuan: the existing fields are still there but no longer used. 

The deprecation notes can be added during a release, it's also in the descriptions of those old fields.

EnriqueL8: I see that, but they should still be used as a backup for a least a release so we do not break customers in a minor version. And then remove them in a later release such as https://github.com/hyperledger/firefly/pull/1289

Chengxuan:
@EnriqueL8 Migration logic added, ready for another look

internal/ethereum/ethereum.go
```diff
@@ -66,12 +66,27 @@ func NewEthereumConnector(ctx context.Context, conf config.Section) (cc ffcapi.A
 		eventBlockTimestamps:       conf.GetBool(EventsBlockTimestamps),
 		eventFilterPollingInterval: conf.GetDuration(EventsFilterPollingInterval),
 		traceTXForRevertReason:     conf.GetBool(TraceTXForRevertReason),
-		retry: &retry.Retry{
-			InitialDelay: conf.GetDuration(RetryInitDelay),
-			MaximumDelay: conf.GetDuration(RetryMaxDelay),
-			Factor:       conf.GetFloat64(RetryFactor),
-		},
+		retry:                      &retry.Retry{},
```
EnriqueL8, line 69: Now that I see this I'm confused, it seems that this was the correct level for the connector.retry for the HTTP and WSS configuration but just because it turns out that it's being used for the block listener then we are completely deprecating the `connector.retry` section? and moving things up to `connector.blockListener`? I think we need to keep both

EnriqueL8 (changes requested):
After looking the PR - I think bigger changes are needed here. So we use these delay and retry configurations across three things:
- Block Listener
- Event Listener
- Event Stream

If we are deciding to split one, we should split them all to be configurable. Right not this PR proposes only splitting block listeners but those configurations still apply to the other two which is odd

Chengxuan:
@EnriqueL8 thanks for the question. Even though the `retry` object appeared in the 3 different concepts you listed above, the loop that it controls all serves the same purpose which is to resume the listener process. So I think it makes sense to put them into one configuration until there is a need to split them.

I'll pick a better name for the configuration to make that clear.

EnriqueL8 (approved):
Makes sense now! Thanks for the clarification @Chengxuan

PR #160, opened 2024-12-11
Ensure the descriptions are represented in HTML nicely

Chengxuan:
The `<deprecated>` string was treated as an HTML element by markdown, therefore hidden.

Here is the fixed version:

image


EnriqueL8 (approved):
LGTM

PR #162, opened 2025-02-24
pick up configuraitons for monitoring and liveness path

Chengxuan:
picks up https://github.com/hyperledger/firefly-transaction-manager/releases/tag/v1.3.19


EnriqueL8 (approved):
Looks good!

Due to an issue with the goheader tool it believes that those copyright files should be updated to 2025 and that is why they have been modified as part of this PR

PR #165, opened 2025-03-13
Update common and signer for release

EnriqueL8:
Updating common and signer to latest for release

Had to fix the block listener as well to always check if the context is cancelled before the timeout, this is because some requests might take a long time and just runs infinitely 

You can clearly see the issue that it was hitting https://github.com/hyperledger/firefly-evmconnect/actions/runs/13833748049/job/38703773942?pr=165#step:3:1068


matthew1001 (approved):
LGTM

PR #176, opened 2025-10-29
Expose ws rpc

Chengxuan:
Adding another function `WSRPC()` to expose the websocket RPC client.

> NOTE: the websocket RPC client is optional, so the client should check whether it's nil

The HTTP RPC client is already exposed by the `RPC()` function.

### DIFF
--- a/internal/ethereum/blocklistener.go
+++ b/internal/ethereum/blocklistener.go
@@ -27,7 +27,6 @@ import (
 	"github.com/hyperledger/firefly-common/pkg/fftypes"
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
-	"github.com/hyperledger/firefly-common/pkg/wsclient"
 	"github.com/hyperledger/firefly-evmconnect/internal/msgs"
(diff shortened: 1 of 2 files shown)


internal/ethereum/blocklistener.go
```diff
@@ -88,9 +88,7 @@ func newBlockListener(ctx context.Context, c *ethConnector, conf config.Section,
 		unstableHeadLength:         int(c.checkpointBlockGap),
 		hederaCompatibilityMode:    conf.GetBool(HederaCompatibilityMode),
 	}
-	if wsConf != nil {
-		bl.wsBackend = rpcbackend.NewWSRPCClient(wsConf)
-	}
+
```
EnriqueL8: I'm confused why the WS client cannot still be initialized here?

EnriqueL8 (commented):
Thanks @Chengxuan - one question, seems weird to create a block listener that doesn't actually work because the client is nil.

internal/ethereum/blocklistener.go
Chengxuan: @EnriqueL8 WS is optional. The connector doesn't require WS to work.

EnriqueL8 (approved):
Looks good

PR #187, opened 2026-03-23
Pure ethrpc interface on ethblocklistener rather than hybrid FFCAPI

peterbroadhurst:
We ended up between two stools on our `ReconcileConfirmationsForTransaction` API, where if you used it directly via `ethblocklistener` you were still exposed to some types from `ffcapi`.

This PR fully cleans up the separation.

### DIFF
--- a/internal/ethereum/ethereum.go
+++ b/internal/ethereum/ethereum.go
@@ -37,6 +37,7 @@ import (
 	"github.com/hyperledger/firefly-evmconnect/pkg/ethblocklistener"
 	"github.com/hyperledger/firefly-evmconnect/pkg/ethrpc"
 	"github.com/hyperledger/firefly-signer/pkg/abi"
+	"github.com/hyperledger/firefly-signer/pkg/ethtypes"
 	"github.com/hyperledger/firefly-signer/pkg/rpcbackend"
 	"github.com/hyperledger/firefly-transaction-manager/pkg/ffcapi"
 )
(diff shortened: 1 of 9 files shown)


pkg/ethblocklistener/blocklistener.go
```diff
@@ -38,6 +38,24 @@ import (
...
+// the list is sorted by block number
+//   - the first block is the block that contains the transaction hash
+//   - the last block is the most recent confirmation
+//
+// this list can be used as input to the future reconcile request to avoid re-fetching the blocks if they are no longer
+// in the in-memory partial chain
+// WARNING: mutation to this list is not expected, invalid modifications will cause inefficiencies in the reconciliation process
+//
+//	`rebuilt` will be true if an invalid confirmation list is detected by the reconciliation process
+type ConfirmationUpdateResult struct {
```
Chengxuan, line 51: This seems to be a duplication of the existing ffcapi struct. Is the old one still in use after this change?

internal/ethereum/ethereum.go
```diff
@@ -200,12 +202,51 @@ func withDeprecatedConfFallback[T any](conf config.Section, getter func(string)
 
 // ReconcileConfirmationsForTransaction is the public API for reconciling transaction confirmations.
 // It delegates to the blockListener's internal reconciliation logic.
-func (c *ethConnector) ReconcileConfirmationsForTransaction(ctx context.Context, txHash string, existingConfirmations []*ffcapi.MinimalBlockInfo, targetConfirmationCount uint64) (*ffcapi.ConfirmationUpdateResult, error) {
+func (c *ethConnector) ReconcileConfirmationsForTransaction(ctx context.Context, txHash string, existingConfirmations []*ffcapi.MinimalBlockInfo, targetConfirmationCount uint64) (res *ffcapi.ConfirmationUpdateResult, err error) {
```
Chengxuan, line 205: this is still using ffcapi structs, is this expected?

peterbroadhurst: Yes, this is the FFCAPI interface. I'm actually not sure if there are consumers of it, but I did not want to remove it as it was added as a part of FFCAPI in previous commits.

You'll see it's just a shim mapping from the `ethblocklistener`/`ethrpc` interfaces.

If you directly load `ethblocklistener` through the `pkg` interface introcued in #180 you will never have to see the FFCAPI generic mapping structure.

This is recommended for EVM specific code using `firefly-evmconnect` as a module, rather than a connector building entirely within the FFTM framework.

pkg/ethblocklistener/blocklistener.go
peterbroadhurst: Comments on this below.

internal/ethereum/ethereum.go
Chengxuan: Thanks for the clarification @peterbroadhurst . In this case, we need to decide whether to maintain 2 sets of public interfaces or to mark one of them deprecated. 

Also, the duplicated structs were not direct copy pasted from the ffcapi, ConfirmationUpdateResult no longer has receipt (moved as a separate return value), MinimalBlockInfo no long accept any hashes, it only accepts hex string only (compatible with ethereum's hash choice).

^^ this difference should be documented

Chengxuan (approved):

PR #188, opened 2026-04-20
[blocklistener] Trusted Mode for Minimal JSONRPC Access

onelapahead:
The current block listener is robust, and designed to verify the integrity of the chain head, detecting re-orgs so that it does not confirm an event or transaction before the connector is confident that the blocks are stable/finalized.

For permissioned, BFT chains where finalization is immediate this both unnecessary/inefficient, and requires full JSONRPC access to all block receipts. This PR proposes an alternative mode that minimizes the JSONRPC footprint, and trusts the `eth_blockNumber` results as the canonical head block number, still using filters/WS heads to wake up on when to poll for the latest block height.

This mode is considered experimental.

### DIFF
--- a/config.md
+++ b/config.md
@@ -94,6 +94,12 @@
 |password|Password|`string`|`<nil>`
 |username|Username|`string`|`<nil>`
 
+## connector.blockListener
+
+|Key|Description|Type|Default Value|
+|---|-----------|----|-------------|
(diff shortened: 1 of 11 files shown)


peterbroadhurst:
Superseded by https://github.com/hyperledger/firefly-evmconnect/pull/190

PR #190, opened 2026-04-21
introduce a light weight mode for tracking block progression

Chengxuan:
for https://github.com/hyperledger/firefly-evmconnect/issues/191

Requires https://github.com/hyperledger/firefly-transaction-manager/pull/159

EVM implementation of the `headBlockNumber` tracking mode. In this new mode, new block hashes event are still used as a shoulder tap event to trigger the chain head refresh using `eth_blockNumber` call, which will fetch the latest chain head. However, the logic no longer makes the heavy calls to fetch the detailed information of each block; therefore, it hugely reduces the amount of data fetched from the configured JSON-RPC endpoint.

The following eth method is no longer required when `headBlockNumber` mode is used for tracking block progression.

- `eth_getBlockByHash`
- `eth_getBlockByNumber`


README.md
```diff
@@ -34,40 +34,63 @@ For a full list of configuration options see [config.md](./config.md)
...
 - `eth_sendTransaction`
 - `eth_getTransactionCount`
 - `eth_sendRawTransaction`[^2]
 
-
 [^1]: also used by Transaction submission if the handler is configured to get gas price using "connector".
 
 [^2]: only required by custom transaction handlers that supports pre-signing.
+
+### Optional methods
```
hosie, line 72: we use footnotes to mark `eth_sendRawTransaction` as optional but we now have a whole other  section for optional methods.   Would it make sense to move `eth_sendRawTransaction` to the `Optional methods` section now?

pkg/ethblocklistener/confirmation_reconciler.go
```diff
@@ -40,6 +41,32 @@ func toBlockInfoList(ffcapiBlocks []*ethrpc.MinimalBlockInfo) (blocks []*ethrpc.
...
+			// the transaction is not yet included in the chain head
+			return nil, nil, i18n.NewError(ctx, msgs.MsgTransactionNotIncludedInChainHead, txHash, chainHead, txReceipt.BlockNumber.String())
+		}
+		currentConfirmationCount := chainHead - txReceipt.BlockNumber.Uint64()
+		if currentConfirmationCount > targetConfirmationCount {
+			currentConfirmationCount = targetConfirmationCount
+		}
+		return &ConfirmationUpdateResult{
+			Confirmed: currentConfirmationCount == targetConfirmationCount,
+			// no support on fork detection in this mode
```
hosie:
```suggestion
			// no support on fork detection in this mode from the connector although firefly transaction manager will inject `NewFork: true` if it detects a reduction in confirmationCount. 
```

hosie (commented):
LGTM with 2 comments

README.md
Chengxuan: good call, will move that over

internal/ethereum/config.go
```diff
@@ -19,13 +19,15 @@ package ethereum
...
 	"github.com/hyperledger/firefly-common/pkg/wsclient"
+	"github.com/hyperledger/firefly-transaction-manager/pkg/ffcapi"
 )
 
 const (
 	ConfigGasEstimationFactor   = "gasEstimationFactor"
 	ConfigDataFormat            = "dataFormat"
 	BlockPollingInterval        = "blockPollingInterval"
 	BlockCacheSize              = "blockCacheSize"
+	BlockTrackingMode           = "blockTrackingMode"
```
peterbroadhurst: Per https://github.com/hyperledger/firefly-transaction-manager/pull/159 feedback and discussion, I believe the updated spelling for this configuration properly is:

`confirmationMode` : `light`|`full`

Where a good description in the `config.md` could be:

- The level of verification performed within the connector to establish confirmation. In light mode the block number alone is used. In full mode the full chain of blocks is verified in the connector, and attaached to receipts and events.

Within the code, this means:

- `light` mode needs indicate to the FFTM framework that the confirmation manager (in FFTM) should only ask the EVM code about the latest block numbers
- `full` node needs to indicate to the FFTM framework that the confirmation manager (in FFTM) should ask the EVM code about the full details of blocks

The fact the configuration is in EVMConnect, but the change of behavior is in FFTM, means that the FFCAPI change is (a little confusingly) `BlockListenerTrackingMode` with two options `InMemoryPartialChain` and `HeadBlockNumber`.

> I note the alternative would be place the `confirmationMode` setting directly in FFTM under `config.confirmations` where the actual code that operates the confirmation manager runs, in which case there would be simplified spelling.

Chengxuan: @peterbroadhurst thanks for the breakdown. 

> I note the alternative would be place the confirmationMode setting directly in FFTM under config.confirmations 

I have the same concern. How about `chainTrackingMode` or `chainSyncMode` for the connector configuration?


Here is a more detailed explaination of the two modes:


`xxxxxxxMode: light|full`

Where a good description in the `config.md` could be:

The level of block information tracked / synchronised within the connector to establish confirmation and support block indexing. 
- In `full` mode the full partial chain of blocks is verified in the connector, and attached to receipts and events. 
- In `light` mode the block number alone is used, and block metadata are not fecthed by the connector. Compared to full mode, this mode:
  - No block-level information for confirmations
  - No support of block listener
  - Event confirmation is numeric depth
  - Reorg detection is limited (no detection until final confirmation)

Within the code, this means:

- `light` mode: 
  - Confirmation logic needs to be aware of this mode and expect it will only be given the latest block number, not block metadata. There are user-facing 2 consumers needs to be updated: 1. `ReconcileConfirmationsForTransaction` function in the connector, 2. confirmation manager in FFTM
  - Block indexing needs to be aware of this mode, as it cannot function under this mode. One user-facing consumer is FFTM's block listener.
- `full` node
  - Confirmation logic needs to be aware of this mode and expect it to be given a locally consistent partial chain with detailed block metadata. Therefore, a more robust confirmation output can be provided using this locally consistent partial chain. There are 2 consumers needs to be updated: 1. `ReconcileConfirmationsForTransaction` function in the connector, 2. confirmation manager in FFTM
  - Block indexing logic can function normally in this mode.

Given confirmation behaviour change is just one downstream effect, I feel `chainTrackingMode` or `chainSyncMode` are better names.

peterbroadhurst (approved):
Marking approval, as @hosie and @Chengxuan discussed this and reached a joint reasoned proposal on the spelling of:

- The setting is `connector.chainTrackingMode`, and is enabled on the the specific connector implementation (`firefly-evmconnect` in this case)
- The two possible values are:
    - `full`
    - `light`

These values are described in code comments on the FFTM side:

https://github.com/kaleido-io/firefly-transaction-manager/blob/852459b5a29d758290410fb27617a66d6fd28cf6/pkg/ffcapi/api.go#L104-L107

Chengxuan:
Since https://github.com/hyperledger/firefly-evmconnect/pull/190/changes/457132d05b89d15a23fc13f70e4e3028a27f7993 contains only removals of existing contents in the Readme, and I don't have access to the Snyk logs yet. I'm going to merge the PR despite the Snyk failure

PR #192, opened 2026-05-01
Limit estimate gas to the current blockLimit

peterbroadhurst:
It is possible to get into a problematic situation where a transaction is accepted into the EVMConnect transaction queue, with a gas estimation that is higher than the current block limit.

This is because the `eth_estimateGas` configuration on the node-side does not use the current `gasLimit` as the limit on the call. For example in the default Besu configuration, the `--rpc-gas-cap` option can be unlimited (`0`) or 100M. Whereas the block `gasLimit` is only 4M by default.

This PR proposes a best-effort protection on this, when the block tracking mode is full and EVMConnect is monitoring the highest blocks from the blockchain. It tracks the current `gasLimit` from the head of the chain, and if a transaction gas estimate (after scaling) breaches this limit then an error is thrown.

Because the gas estimate occurs before the transaction is accepted, this rejects the transaction with a suitable error early and prevents allocation of a nonce.


Chengxuan (approved):
LGTM

PR #194, opened 2026-05-13
Simplify the blocklistener_test.go with wrappers on mocks

peterbroadhurst:
The longhand `mRPC.On("CallRPC"` in `blocklistener_test.go` makes it really hard to read the tests, particularly due to the fact that the args have to be set in a `Run()` due to the nature of the RPC interface.

So this PR proposes no functional changes, just an introduction of wrappers to make the code more readable and easier to read and diagnose.


peterbroadhurst:
Merging this into https://github.com/hyperledger/firefly-evmconnect/pull/193 now I've checked through it in isolation and seen a green build.

PR #197, opened 2026-06-25
add method for checking eip1559 support

Chengxuan:
Added a method for consumers to get the block info of the highest block. So that it can be used to check whether EIP1559 is supported using the presence of "baseFeePerGas" field from the block header.

Note: instead of introducing a `GetBlockBaseFeePerGas` or `SupportsEIP1559` in the BlockListener interface. I decided to cache and return the entire block info. So if other fields in the block info need to be used for other checks, we don't need to keep adding separate functions for them at the block listener interface.


peterbroadhurst (approved):
👍 - looks clear and straight forward.
The `BlockInfoJSONRPC` doesn't contain the transactions, so the extra memory is trivial.

PR #198, opened 2026-06-26
chore: migrate to hyperledger-firefly org, firefly deps, Go 1.26.0

dan-r:
Releases needed for signer and transaction-manager. Currently just pointing @ main


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.23-alpine3.20 AS builder
+FROM golang:1.26.4-alpine3.24 AS builder
```
onelapahead: I don't recommend pinning the patch version for Go builds, it ensures we patch CVEs whenever we build

go.mod
```diff
@@ -1,13 +1,13 @@
-module github.com/hyperledger/firefly-evmconnect
+module github.com/hyperledger-firefly/evmconnect
 
-go 1.23.0
+go 1.26.4
```
onelapahead: I think its best to avoid pinning to specific patches `go 1.26.0` is still compatible with newer Go 1.26 versions, but if a project importing this was on an older go 1.26 (for whatever reason) we force it to upgrade.

.github/workflows/go.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: "1.23"
+          go-version: "1.26.4"
```
onelapahead: same here - just 1.26

go.mod
EnriqueL8: done

Chengxuan (approved):
LGTM. 

Github action seems to be broken due to https://github.com/hyperledger-firefly/common/pull/225 is required.

Dockerfile
EnriqueL8: argh I pinned it

PR #201, opened 2026-07-06
Adding an option to cache txn receipts when building canonical chain

Chengxuan:
Adding an option to fetch txn receipts for each new block when building the canonical chain.

When this is turned on, it will optimize the efficiency of fetching transaction receipts for the consumers to use.

This comes with a trade-off that transaction receipts that are not needed will still be downloaded.

Combining with  `connector.useGetBlockReceipts: true`, it can significantly reduce the number of JSON-RPC calls for blocks with a large amount of transactions that the consumer requires receipts for.

### DIFF
--- a/pkg/ethblocklistener/blocklistener.go
+++ b/pkg/ethblocklistener/blocklistener.go
@@ -62,6 +62,8 @@ type BlockListenerConfig struct {
 	BlockPollingInterval          time.Duration            `json:"blockPollingInterval"`
 	HederaCompatibilityMode       bool                     `json:"hederaCompatibilityMode"`
 	BlockCacheSize                int                      `json:"blockCacheSize"`
+	ReceiptCacheEnabled           bool                     `json:"receiptCacheEnabled"`
+	ReceiptCacheSize              int                      `json:"receiptCacheSize"`
 	IncludeLogsBloom              bool                     `json:"includeLogsBloom"`
 	UseGetBlockReceipts           bool                     `json:"useGetBlockReceipts"`
 	MaxAsyncBlockFetchConcurrency int                      `json:"maxAsyncBlockFetchConcurrency"`
@@ -121,6 +123,7 @@ type blockListener struct {
 	consumerMux                   sync.Mutex // covers consumers and listenLoopDone
 	consumers                     map[fftypes.UUID]*BlockUpdateConsumer
 	blockCache                    *lru.Cache
+	txReceiptCache                *lru.Cache
 	blockFetchConcurrencyThrottle chan *blockReceiptRequest
 	BlockListenerConfig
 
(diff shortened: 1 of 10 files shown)


pkg/ethblocklistener/blocklistener.go
```diff
@@ -520,7 +533,9 @@ func (bl *blockListener) handleNewBlock(mbi *ethrpc.BlockInfoJSONRPC, addAfter *
 
 	// Ok, we can add this block
 	var newElem *list.Element
+	forkTrim := false
 	if addAfter == nil {
+		bl.resetReceiptCache()
```
peterbroadhurst, line 539: If I've understood correctly, this is quite a blunt instrument for cache cleanup.

Let's say we have 100 blocks of history, and we have a (potentially 1000s) of receipts in that cache.
The re-org is likely to affect just the last block or two - but here we will abandon the entire receipt cache and move onto a new generation.

Given consumers of this library are very unlikely to be listening at head for chains with re-orgs, and rather a long way back, this feels potentially likely to make the cache completely redundant.

So it feels like this feature currently is shaped only to be really helpful in chains where re-orgs are uncommon.

To make the cache work in such chains, you'd probably have to have a separate structure in the cache with the list of receipts for each block - so you could clear blocks after the fork point. So I understand that's a big increase in complexity.

As such - feel free to take this as something to factor in when considering if the feature is useful, rather than necessarily a requirement to change the feature in this round.

peterbroadhurst (approved):
Noting one current limitation of the applicability of the feature in its current form to chains with high re-org likelihood.

However, it's a suggestion for a future potential improvement rather than a blocker.

pkg/ethblocklistener/blocklistener.go
Chengxuan: Yes. That's correct. I'll add a comment there to state this is a deliberate choice.

peterbroadhurst: ... as I think about it, I wonder if the processing of just selective clearing of the cache with block < X, which would be very trivial to add vs. just the `.Purge()`.

While it's O(N), it's extremely straight forward and the cost of refetching receipts with I/O (on both sides of the wire) is high.

Chengxuan: Agreed. Fork trim is easier to implement because we'll know the ones that are invalid. The trade off is the delayed revocation of invalid txns in earlier blocks which should just be locking delays and minimal

Rebuilding the canonical chain path is harder.

Chengxuan: @peterbroadhurst Here is the follow up PR to improve the efficiency of cache invalidation for fork trim: https://github.com/hyperledger-firefly/evmconnect/pull/206

PR #202, opened 2026-07-07
Fix docker release workflow

dan-r:
Derive docker release image tags from release event tag

### DIFF
--- a/.github/workflows/docker_release.yml
+++ b/.github/workflows/docker_release.yml
@@ -32,22 +32,22 @@ jobs:
       - name: Set latest tag
         if: github.event.action == 'released'
         run: |
-          echo "DOCKER_TAGS=${{ env.DOCKER_TAGS }},ghcr.io/${{ github.repository }}:latest" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_release.yml
```diff
@@ -32,22 +32,22 @@ jobs:
       - name: Set latest tag
         if: github.event.action == 'released'
         run: |
-          echo "DOCKER_TAGS=${{ env.DOCKER_TAGS }},ghcr.io/${{ github.repository }}:latest" >> $GITHUB_ENV
+          echo "DOCKER_TAGS=,ghcr.io/${{ github.repository }}:latest" >> $GITHUB_ENV
```
EnriqueL8: There is an extra `,` at the beginning

EnriqueL8 (commented):
Got it, one fix on the DOCKER_TAGS

EnriqueL8 (approved):
Thanks - so we are simplifying this to not allow multiple tags on alpha, beta and rc which makes sense

PR #173, opened 2025-08-13
fixing nil pointer erorr in the filter enrichment logic

Chengxuan:
resolves #172

PR #203, opened 2026-07-08
fix: restore group 0 as primary group for runtime user

EnriqueL8:
PR #173 added an explicit passwd entry for uid 1001 with primary group 1001 (evmgroup). Previously the bare 'USER 1001' had no passwd entry, so the runtime user's primary group defaulted to 0 (root).

The FireFly CLI initializes the evmconnect data volume with 'chgrp -R 0 && chmod -R g+rwX' (the group-0 convention, matching the ADD/COPY --chown=1001:0 lines in this Dockerfile), so a runtime user outside group 0 can no longer write to it. Every release since v1.4.1 fails against a CLI-created stack with:


Chengxuan (approved):
LGTM Thanks for the fix @EnriqueL8

PR #216, opened 2026-08-17
feat(blocklistener): Block height metrics for canonical chain vs target node

onelapahead:
To help to detect if either the underlying node infrastructure or the block listener itself is unreliable / out-of-sync, we add two new gauges:
* the "target" block height - what we observe on an independent monitor go routine using `eth_blockNumber`
* the "canonical" block height - when we're building a canonical chain via block filters / newHeads subscriptions (when in full chain tracking mode), we want to know the height of that chain.

Using such metrics we can _alert_ on several different scenarios:
1. is the target block height not incrementing - depending on the underlying consensus/EVM implementation, this might indicate the node is failing to sync the chain.
2. are the block listener canonical height and the target height out of sync within some toleration/expectation - this would either indicate the endpoint has a bad filter/subscription and is failing to notify the listener of new blocks, or the listener itself is somehow failing to track the chain otherwise.
3. if we are able to monitor the chain height itself via other means (a synced node we're monitoring, a block explorer, etc.), we can detect the difference of the target/canonical with the monitor to detect any lag.

### DIFF
--- a/pkg/ethblocklistener/blocklistener.go
+++ b/pkg/ethblocklistener/blocklistener.go
@@ -313,8 +313,7 @@ func (bl *blockListener) establishBlockHeightWithRetry() error {
 }
 
 // queryBlockHeightFromRPC queries eth_blockNumber and returns the result, without updating any listener
-// state. The height the node reports is recorded on the target block height gauge - the only source of
-// that metric, so it is always exactly what the node last told us.
+// state. Caller must not hold canonicalChainLock. The height the node reports is recorded on the target block height gauge.
 func (bl *blockListener) queryBlockHeightFromRPC() (uint64, error) {
 	var hexBlockHeight ethtypes.HexInteger
 	rpcErr := bl.backend.CallRPC(bl.ctx, &hexBlockHeight, "eth_blockNumber")
@@ -376,10 +375,8 @@ func (bl *blockListener) listenLoop() {
 			}
 		}
 
-		// In full chain tracking mode the loop below never queries the height the node reports, so refresh
-		// it here for the target metric. Done ahead of the filter calls, so that when they start failing -
-		// exactly when the gap between the node and the chain we have built matters - we can still see the
-		// chain moving on without us. No-ops (and makes no query) when metrics are not enabled.
+		// In full chain tracking mode, the loop below never queries the height the node reports, so we refresh
+		// it here for the target metric. Done ahead of the filter calls.
 		if bl.ChainTrackingMode != ffcapi.ChainTrackingModeLight {
 			bl.refreshTargetBlockHeightMetric()
 		}
@@ -800,11 +797,11 @@ func (bl *blockListener) GetHeadBlockNumber(_ context.Context) uint64 {
 }
 
 func (bl *blockListener) setHighestBlock(block uint64) {
+	defer bl.setBlockHeightMetric(metricCanonicalBlockHeight, block)
 	bl.canonicalChainLock.Lock()
+	defer bl.canonicalChainLock.Unlock()
 	bl.highestBlock = block
 	bl.highestBlockSet = true
-	bl.canonicalChainLock.Unlock()
-	bl.setBlockHeightMetric(metricCanonicalBlockHeight, block)
 }
 
 // checkAndSetHighestBlock records the chain head height and caches full block info for the head.
@@ -817,8 +814,7 @@ func (bl *blockListener) checkAndSetHighestBlock(bi *ethrpc.BlockInfoJSONRPC) {
 		bl.highestBlockSet = true
 		bl.headBlockInfo = bi
 		// The gauge is bound to the same variable GetHighestBlock reports to event streams, so it is the
-		// head we are actually tracking rather than a separate sample of it. Just a gauge set, no I/O,
-		// so we accept doing it while we hold the write lock.
+		// head we are actually tracking rather than a separate sample of it.
 		bl.setBlockHeightMetric(metricCanonicalBlockHeight, block)
 	} else if block == bl.highestBlock {
 		// Height already known from eth_blockNumber. Store the first full block at that height.
--- a/pkg/ethblocklistener/blocklistener.go
+++ b/pkg/ethblocklistener/blocklistener.go
@@ -414,8 +414,7 @@ func (bl *blockListener) listenLoop() {
 				continue
 			}
 			// In light mode there is no canonical chain being built, so the head we dispatch to
-			// consumers is what we report as the canonical height (the metrics loop separately
-			// reports the target height)
+			// consumers is what we report as the canonical height
 			if head == bl.currentChainHead {
 				failCount = 0
 				continue
--- a/pkg/ethblocklistener/blocklistener.go
+++ b/pkg/ethblocklistener/blocklistener.go
@@ -314,8 +314,8 @@ func (bl *blockListener) establishBlockHeightWithRetry() error {
 	})
 }
 
-// refreshHighestBlockFromRPC updates highestBlock from eth_blockNumber. Caller must not hold canonicalChainLock.
-func (bl *blockListener) refreshHighestBlockFromRPC() (uint64, error) {
+// queryBlockHeightFromRPC queries eth_blockNumber and returns the result, without updating any listener state.
+func (bl *blockListener) queryBlockHeightFromRPC() (uint64, error) {
 	var hexBlockHeight ethtypes.HexInteger
 	rpcErr := bl.backend.CallRPC(bl.ctx, &hexBlockHeight, "eth_blockNumber")
 	if rpcErr != nil {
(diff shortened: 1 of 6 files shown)


pkg/ethblocklistener/blocklistener_metrics.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// The target height requires a query of the node
+	head, err := bl.refreshHighestBlockFromRPC()
+	if err != nil {
+		// Purely a metrics query - the listen loop has its own error handling for the chain state
+		log.L(bl.ctx).Warnf("Failed to query target block height for metrics: %s", err)
+		return
+	}
+	bl.setBlockHeightMetric(metricTargetBlockHeight, head)
+}
```
Chengxuan: @onelapahead can you elaborate on the thinking behind here? The metrics-emitting logic does functional logic (updates the chain head), rather than purely capturing metrics.

Have you considered making the metrics-emitting logic do a no-op or emit an "unavailable/error" status metric when the main block height setting loop is in an unready state?

Chengxuan (commented):
@onelapahead thanks for adding metrics. Agree with the intention of detecting out-of-sync / unreilable block listener head update. However, adding block head update logic in the metrics logic voilates single responsibilty practice, which make the logic more complex and bugs in the block listener logic harder to spot.

pkg/ethblocklistener/blocklistener_metrics.go
onelapahead: That is not the case nor intent - `getCanonicalBlockHeight()` acquires a RLock to read the listener's state but otherwise does not update anything as I understand it.

I've renamed `refreshHighestBlockFromRPC()` bc that was misleading, its named `queryBlockHeightFromRPC` and its just purely an `eth_blokcNumber` call, but it doesn't change any listener state.

pkg/ethblocklistener/blocklistener.go
```diff
@@ -400,17 +407,20 @@ func (bl *blockListener) listenLoop() {
 		}
 
 		if bl.ChainTrackingMode == ffcapi.ChainTrackingModeLight {
```
peterbroadhurst, line 418: Can I just check the metric emission hasn't been made specific to light mode

onelapahead: No - no matter what the metrics are emitted in all modes.

What you're seeing is a metric behavior difference between light mode and full mode - bc light mode does not track a canonical chain, the height it reports for the canonical chain height is the RPC's own block height.

pkg/ethblocklistener/blocklistener_metrics.go
Chengxuan: @onelapahead  Thanks for the renaming. 

So what was added in the PR is a separate goroutine that
1. runs on the same timer as the main listener goroutine.
2. doing a separate & extra  `eth_blockNumber` call on the same timer


That approach has its merit, which is very easy to understand as a small patch, but doesn't feel like the most efficient way of capturing the diagnostic information. Also, lacks the next level of granularity to pin down the problem for diagnosis, because the metrics are not associated with any variable that's driving the functional logic.


Breaking down the situation you are trying to resolve, I'd give the following suggestions:

1. is the target block height not incrementing (node out of sync) - depending on the underlying consensus/EVM implementation, this might indicate the node is failing to sync the chain.

  > it's a node running problem, metrics and alert should be added in the JSON-RPC node.

2. are the block listener canonical height and the target height out of sync within some toleration/expectation - this would either indicate the endpoint has a bad filter/subscription and is failing to notify the listener of new blocks, or the listener itself is somehow failing to track the chain otherwise.

> metrics should be added to track those failures. e.g. `checkAndSetHighestBlock` (and/or`queryBlockHeightFromRPC`) should emit failure metrics and gauge metrics for new head, so when the head stays stale, it's obvious. It can be used as a trigger to escalate to the JSON-RPC node provider ops to check their metrics as well if the metric is from `queryBlockHeightFromRPC`.

4. if we are able to monitor the chain height itself via other means (a synced node we're monitoring, a block explorer, etc.), we can detect the difference of the target/canonical with the monitor to detect any lag.

> this  should be achieveable by adding metrics to `checkAndSetHighestBlock` and `queryBlockHeightFromRPC` 



So I'd suggest replacing the new goroutine with metrics in `checkAndSetHighestBlock` and `queryBlockHeightFromRPC`. If you feel there is a need to do `queryBlockHeightFromRPC` in full-tracking mode, we could think about where to add it in the existing logic to improve the function logic. (Suggest a separate PR for this to make the current PR single-purposed)

onelapahead: > it's a node running problem, metrics and alert should be added in the JSON-RPC node.

Be mindful - EVM connect user don't always own the JSON/RPC node or endpoint they are connecting to. So the `target_chain_height` metric is very much meant to be a sentinel for such use cases. We have observed users connecting to unstable/questionable infrastructure from time to time.

> metrics should be added to track those failures

Great suggestion - I agree. I added failure counters for all the RPC methods we poll in the listener (deliberately not touching other packages).

However, we did observe a filter continually come back empty for an endpoint, but when TM restarted and saw a a much higher block height than what the canonical was tracking, and it went into catch up mode instead. So (although _older_ I'm willing to bet) clients have the potential to "dead air" a filter potentially. Thats where have the `target` vs `canonical` can be very helpful to see in the metrics and logs as a sanity check.

I've incorporate your other feedback as well:
1. Got rid of the separate Goroutine, for full mode, we poll `eth_blockNumber` at the top of the listener loop. Light mode was always polling it. Whenever we poll it via `queryBlockHeightFromRPC`, we set `target_chain_height`. But currently, in full mode we don't do anything different yet if we notice a larger difference in that block height reported by the endpoint, and our listener's `highestBlock`. I think @peterbroadhurst will be making changes to help there instead potentially.
2. I moved `canonical_chain_height` emissions to `checkAndSetHighestBlock` and `setHighestBlock` funcs, to track what the subscription/filter is telling us is the highest block at any point.

pkg/ethblocklistener/blocklistener.go
onelapahead: I've refactored this slightly based on Cheng's feedback (detailed in https://github.com/hyperledger-firefly/evmconnect/pull/216#discussion_r3876898449 above), so hopefully its clearer and simpler now.

peterbroadhurst (approved):
Thanks for the updates, very clear now.

pkg/ethblocklistener/blocklistener_metrics.go
```diff
@@ -0,0 +1,92 @@
...
+		return
+	}
+	mm.IncCounterMetricWithLabels(bl.ctx, metricPollFailures, map[string]string{metricLabelPollFailures: method}, nil)
+}
+
+// refreshTargetBlockHeightMetric queries the node for the height it reports, purely so the target gauge
+// stays current. Only needed in full chain tracking mode.
+func (bl *blockListener) refreshTargetBlockHeightMetric() {
+	if bl.getMetrics() == nil {
+		return // never drive any query of the node when metrics are not enabled
```
Chengxuan, line 85:
```suggestion
		return // never drive any query of the node when the metrics registry is not set
```

Chengxuan: This highlights a problem I overlooked as part of https://github.com/hyperledger-firefly/transaction-manager/pull/160

Since the raw MetricsRegistry interface from ff-common was exposed, it didn't include information about whether the metrics are enabled in the config.

The control was wrapped in the internal Metrics interface: https://github.com/hyperledger-firefly/transaction-manager/blob/162a945b953601c8ea0811dc701a4d4e06d348da/internal/metrics/metrics.go#L90.

But feels like a separate PR to sort this out, as there is a feature gap in the metrics registry

pkg/ethblocklistener/blocklistener_metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	mm, err := registry.NewMetricsManagerForSubsystem(ctx, metricsSubsystem)
+	if err != nil {
+		return i18n.WrapError(ctx, err, msgs.MsgMetricsInitFail, metricsSubsystem)
+	}
+	mm.NewGaugeMetric(ctx, metricTargetBlockHeight, "The block height reported by the connected node via eth_blockNumber", false)
+	mm.NewGaugeMetric(ctx, metricCanonicalBlockHeight, "The block height of the head of the chain tracked by the block listener", false)
+	mm.NewCounterMetricWithLabels(ctx, metricPollFailures, "The number of block listener JSON/RPC polls that have failed, by method", []string{metricLabelPollFailures}, false)
+
+	bl.metricsLock.Lock()
+	defer bl.metricsLock.Unlock()
```
Chengxuan, line 54: The lock is needed because the lifecycle of a metrics registry is tied to a transaction manager, but the block listener is not initiated inside the transaction manager.

In order to avoid a lock, a standalone metrics registry needs to be created first and then passed into both the listener and the transaction manager. That approach will reflect the fact that transaction manager no longer controls the metric prefix for components that emit custom metrics into the metrics that are linked to its metrics endpoint.

Chengxuan (approved):
Updates look good to me.

pkg/ethblocklistener/blocklistener_metrics.go
Chengxuan: Raised https://github.com/hyperledger-firefly/common/issues/245

PR #220, opened 2026-08-27
Start in catchup=true and guard forward only HWM advance

peterbroadhurst:
There is a small window during adding a new, where we set `catchup: false` (meaning we're in the head group), then drop the mutex before actually performing the start

https://github.com/hyperledger-firefly/evmconnect/blob/1e5d5cc2becbaa4ba4457bbaeff87283f65130fe/internal/ethereum/event_stream.go#L163-L164

https://github.com/hyperledger-firefly/evmconnect/blob/1e5d5cc2becbaa4ba4457bbaeff87283f65130fe/internal/ethereum/event_actions.go#L133-L138

There is an small window where if the head group moves the checkpoint forwards before we get to writing our first checkpoint.

This PR:

1. Makes the default `catchup: true`
2. Defers setting `es.updateCount++` to update the catchup group until `startEventListener`
3. Tightens up locking around `hwm` to make sure it's not read outside of a mux, and only moves forwards
4. Restores test coverage


internal/ethereum/event_listener.go
```diff
@@ -226,8 +224,11 @@ func (l *listener) filterEnrichEthLog(ctx context.Context, f *eventFilter, metho
 	blockNumber := trimUint64(ethLog.BlockNumber.Uint64())
 	transactionIndex := trimUint64(ethLog.TransactionIndex.Uint64())
 	logIndex := trimUint64(ethLog.LogIndex.Uint64())
-	if blockNumber < l.hwmBlock {
```
peterbroadhurst, line 229: Read outside of lock

internal/ethereum/event_listener.go
```diff
@@ -226,8 +224,11 @@ func (l *listener) filterEnrichEthLog(ctx context.Context, f *eventFilter, metho
...
-		log.L(ctx).Debugf("Listener %s already delivered event '%s' hwm=%d", l.id, getEventProtoID(blockNumber, transactionIndex, logIndex), l.hwmBlock)
+	l.hwmMux.Lock()
```
onelapahead, line 227: `RLock` ?

internal/ethereum/event_listener.go
```diff
@@ -226,8 +224,11 @@ func (l *listener) filterEnrichEthLog(ctx context.Context, f *eventFilter, metho
...
+	hwmBlock := l.hwmBlock
```
onelapahead, line 228: nit:

it could be:
```
  hwmBlock := l.getHWMBlock()
```

where the (read) lock is then acquired/released. Just to avoid these lock-unlock-but-not-deferred code blocks, and use methods to convey what fields must be lock protected.

onelapahead (approved):

internal/ethereum/event_stream.go
```diff
@@ -160,17 +161,24 @@ func (es *eventStream) addEventListener(ctx context.Context, req *ffcapi.EventLi
...
-	if l.catchup && !removed {
+	startCatchupLoop := !readyForLead && !removed && l.catchupLoopDone == nil /* idempotent - do not spawn a second loop */
+	if readyForLead && l.catchup {
+		l.catchup = false
+		es.updateCount++ // we've flipped catchup to false - so this tells the head group to rebuild and include us
+	}
+	if startCatchupLoop {
 		l.catchupLoopDone = make(chan struct{})
+	}
+	es.mux.Unlock()
```
Chengxuan: Should use `defer` for unlock

internal/ethereum/event_stream.go
```diff
@@ -160,17 +161,24 @@ func (es *eventStream) addEventListener(ctx context.Context, req *ffcapi.EventLi
...
+	if startCatchupLoop {
 		go l.listenerCatchupLoop()
 	}
```
Chengxuan, line 181: Should combine the logic into a single `if startCatchupLoop {` condition block

Chengxuan (approved):
Fix itself looks good. some code hygine comments for you to consider

internal/ethereum/event_listener.go
Chengxuan: This has been addressed in 221

Given that the critical section is just an int and a bool, and the access pattern. I don't think an RWMutex justifies the cost right now. Suggest we keep `sync.Mutex` for now.

internal/ethereum/event_stream.go
peterbroadhurst: As we're kicking off a go-routine, I can do it inside the lock. I'll work on a refactor.

internal/ethereum/event_stream.go
peterbroadhurst: I can work on a refactor with functions to scope the `defer`

peterbroadhurst: ... actually, see next comment. No problem with doing the go routine initiation in the lock

Chengxuan (approved):

PR #221, opened 2026-08-28
Prevent a catchup going past the lead group

peterbroadhurst:
## Keep listener catchup behind the lead group

A listener in individual catchup should never advance its HWM past the lead group's position.

On reaching the catchup target, it should just the lead group.

This PR tightens up the code to ensure that's true, through the following changes.

### Catchup could scan into the re-org-unstable window

The catchup now uses `eventStream.catchupCeiling()` on each poll loop to guard the high point to the `headBlock` seen by the lead group.

That means it can never ask the node for logs from blocks that are in the unstable window for a public chain, which is good because that very small edge cause could create false-negative checkpoint above the base of the unstable window (preventing re-detection of blocks in a re-org which is accounted for only in the head group).

### Head position was established inside the unstable window

`preStartProcessing` established `es.headBlock` at the raw chain head, while `leadGroupSteadyState` parks it at `chainHead - checkpointBlockGap`. Startup therefore published a head position one full `checkpointBlockGap` higher than the safe point,.

### Listeners were classified ready-for-lead against an unestablished head

With `headBlock == -1`, `blockGap` is negative, so `blockGap < catchupThreshold` true and a listener added before startup completed would join the lead group immediately. 

### Unsynchronised shared state

Edge race conditions could be affected by these:

- `es.headBlock` was read without the stream lock (an intentional dirty read). Now `atomic.Int64`
- `l.hwmBlock` was read directly, without `hwmMux`, in `leadGroupCatchup`, `leadGroupSteadyState` and `preStartProcessing`
- `preStartProcessing` iterated `es.listeners` without the stream lock


internal/ethereum/event_enricher.go
```diff
@@ -71,8 +71,8 @@ func (ee *eventEnricher) filterEnrichEthLog(ctx context.Context, f *eventFilter,
 			log.L(ctx).Errorf("Failed to set chain ID due to failed to query chain readiness: %+v", err)
 			return nil, matched, decoded, err
 		}
-		if !resp.Ready {
-			log.L(ctx).Errorf("Failed to set chain ID due to the connector is not ready")
+		if !resp.Ready || len(ee.connector.chainID) == 0 {
+			log.L(ctx).Errorf("Failed to set chain ID ready=%t chainID='%s'", resp.Ready, ee.connector.chainID)
```
peterbroadhurst, line 75: Mainly just tightened up this check to close a test coverage gap

Chengxuan:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance (repo has no CLAUDE.md).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Chengxuan (approved):

PR #223, opened 2026-08-30
Implement LastDetected per new EventListenerHWMResponse in FFTM

peterbroadhurst:
Straight forward implementation of https://github.com/hyperledger-firefly/transaction-manager/pull/182 for EVMConnect

> Note this pulls in FFTM v1.5.3


internal/ethereum/event_listener.go
Chengxuan: Catch up event delivery should use `markDetectedAndDispatch` as well

internal/ethereum/event_stream.go
```diff
@@ -524,12 +522,26 @@ func (es *eventStream) dispatchSetHWMCheckExit(ag *aggregatedListener, events ff
 
 }
 
+// markDetectedAndDispatch records the detection point then (importantly afterwards) pushes the event to FFTM
+func (es *eventStream) markDetectedAndDispatch(ag *aggregatedListener, event *ffcapi.ListenerEvent) (exiting bool) {
+	log.L(es.ctx).Debugf("Detected event %s", event.Event)
+	ag.listenersByID[*event.Event.ID.ListenerID].markDetected(event.Checkpoint.(*listenerCheckpoint))
```
Chengxuan, line 530: potential nil pointer exception in `*event.Event.ID.ListenerID`

Chengxuan (changes requested):
Based on the intention, catch up logic need ot mark detected checkpoint as well.

internal/ethereum/event_stream.go
peterbroadhurst: I couldn't find a path here.
It would certainly be a problem if any code reached here with a nil listener ID.

I think the understandable suggestion was defensive coding, but it's hard requirement for to have a listener ID by this point in the code. So it would be a different `panic`.

So I've put a comment to the function that ensures it's set (the only construction path for this struct).

internal/ethereum/event_listener.go
peterbroadhurst: Thank you. Yes 👍  I've applied it here too.

internal/ethereum/event_stream.go
Chengxuan: thanks for checking and confirming this won't be an issue

Chengxuan (approved):
LGTM

PR #224, opened 2026-08-31
Move all RPC calls to single client that chooses WS/HTTP

peterbroadhurst:
EVMConnect has both a HTTP and a WebSocket configuration.

When a WebSocket configuration exists, the block listener connects the WebSocket and then begins to use it.

However, the rest of the listener code continues to use the HTTP connection - meaning we end up with a split where detection of blocks is pinned to a sticky connection over a WebSocket, but the remaining calls happen over HTTP relying on sticky-session HTTP load balancing for calls that require server-side context like `eth_newFilter` / `eth_getFilterChanges`.

This PR introduces a common RPC layer that all aspects of the connector use, which then perform RPC method based selection of whether to send requests down a WebSocket or HTTP connection.

A mix still has benefits:

- High volume submission can utilize concurrent HTTP connections such as `eth_sendRawTransaction`
- Calling read operations like `eth_call` can use multiple backend nodes
- Expensive calls like `debug_traceTransaction` can be moved off of the WebSocket

However, the mix previously used was not ideal.

So the approach here is to have a mode switch that supports:

1. `ws` everything is pushed down the WebSocket once estsablished
2. `http` everything is pushed down HTTP - only use of WebSocket is subscriptions (shoulder taps for new blocks)
3. `auto` a new split that is documented in detail in this PR - the *new default*
4. `legacy` as close as possible to the existing split, with two differences documented in the code

### DIFF
--- a/config.md
+++ b/config.md
@@ -82,6 +82,7 @@
 |maxIdleConnsPerHost|The max number of idle connections, per unique hostname. Zero means net/http uses the default of only 2.|`int`|`100`
 |passthroughHeadersEnabled|Enable passing through the set of allowed HTTP request headers|`boolean`|`false`
 |requestTimeout|The maximum amount of time that a request is allowed to remain open|[`time.Duration`](https://pkg.go.dev/time#Duration)|`30s`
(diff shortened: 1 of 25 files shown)


Chengxuan (approved):
The merge looks great. Thanks for keep the `legacy` option as well.

PR #225, opened 2026-08-31
New mode of head-polling using client-side state management only

peterbroadhurst:
There are cases where a stable WebSocket or sticky-session enabled HTTP load balancer is not available to connect to a bank of RPC nodes for an EVM blockchain.

In these cases relying on the server-side filter management provided by execution clients like Besu/Erigon/Reth/Nethermind etc. via `eth_newFilter` / `eth_getFilterChanges` is impossible, as every HTTP request gets routed to a new node and the filter is constantly lost and recreated.

This PR proposes a new mode of operation where **no server-side state** is required during either catchup or steady-state operation of the block listener. This is supported in both `light` and `full` block tracking modes.

The architecture of the new mode of operation is described in this comment, and I've called out line-level comments on other areas where I've made changes:
https://github.com/hyperledger-firefly/evmconnect/blob/f466460f507ef627ef05b54ef87dd6836ded0b90/internal/ethereum/event_stream_getlogs.go#L175-L220


internal/ethereum/event_listener.go
```diff
@@ -114,7 +114,7 @@ func (l *listener) ensureHWM(ctx context.Context) error {
 			return err
 		}
 		// HWM is the configured fromBlock
-		l.hwmBlock = int64(firstBlock) //nolint:gosec // convert to int64 to match the type of hwmBlock, we should change the type of hwmBlock to uint64
+		l.hwmBlock = blockNumberToInt64(firstBlock)
```
peterbroadhurst, line 117: Given this was getting proliferated, I've condensed to a single place and a single behavior if we ever ended up with the (invalid) case of a block in the >maxint64 range.

internal/ethereum/event_stream.go
```diff
@@ -489,7 +501,13 @@ func (es *eventStream) streamLoop() {
 
 		// We then transition to our steady state, filtering from the front of the chain.
 		// But we might fall behind and need to go back to the catchup mode.
-		if es.leadGroupSteadyState() {
+		var exiting bool
+		if es.c.eventFilterPollingMode == FilterPollingModeClient {
+			exiting = es.leadGroupSteadyStateGetLogs()
+		} else {
+			exiting = es.leadGroupSteadyState()
+		}
```
peterbroadhurst, line 527: Deliberately a hard split here, to protect the existing code path from churn.

internal/ethereum/event_stream.go
```diff
@@ -272,15 +273,26 @@ func (es *eventStream) leadGroupCatchup() bool {
...
+		// Catchup only polls blocks that are outside the re-org unstable window at the head of
+		// the chain (checkpointBlockGap behind the head).
+		// The steady-state loops own delivery of the unstable window.
+		// We stop on the first page where the end lands between catchupThreshold+checkpointBlockGap
+		// (say 550) and the checkpointBlockGap (say 50) before the head to do the switch.
+		pollableHead := blockNumberToInt64(chainHeadBlock) - es.c.checkpointBlockGap
+		if pollableHead < 0 {
+			pollableHead = 0
+		}
+		headGap := pollableHead - fromBlock
```
peterbroadhurst, line 285: This might seem decoupled from the primary feature, but this precision on the right change-over point from catchup to steady-state is more important for client-side filtering (particularly in light mode).

In client-side filtering we just maintain a block number, and page forwards from there. Things behind the earliest block we consume from are completely ignored. So the steady state looks a lot like the catchup mode, but we are extra vigilant to re-detect things in the in the `checkpointBlockGap` (the unstable part).

So it's really important there's no case where this catchup never goes beyond that `checkpointBlockGap`, and before there was an edge case where the last page could land in that unstable window.

Note the window *only* existed in the `leadGroupCatchup` path.

The `listenerCatchupLoop` that is used when a new listener is added that's behind the lead group already had the strong protection against joining the lead group in the `checkpointBlockGap`.

Chengxuan (approved):
The changes to intorduce a client mode looks good to me.

Separate leadGroupSteadyStateGetLogs function looks good to keep the code easier to read.  


@peterbroadhurst I raised 2 follow-on PRs to add some guardrails and enhancements to explicitly resolve delays when the head is behind / rolled back. 

- https://github.com/hyperledger-firefly/evmconnect/pull/225
- https://github.com/hyperledger-firefly/evmconnect/pull/229

PR #226, opened 2026-09-02
Surface block timestamp in confirmationUpdateResult

davecrighton:
Surface the block timestamp in the ConfirmationUpdateResult. Scope to full mode only since we do not get the block info in light mode.

### DIFF
--- a/pkg/ethblocklistener/blocklistener.go
+++ b/pkg/ethblocklistener/blocklistener.go
@@ -50,13 +50,13 @@ import (
 //
 //	`rebuilt` will be true if an invalid confirmation list is detected by the reconciliation process
 type ConfirmationUpdateResult struct {
-	Confirmations            []*ethrpc.MinimalBlockInfo `json:"confirmations,omitempty"`  // the confirmation list
-	Rebuilt                  bool                       `json:"rebuilt,omitempty"`        // when true, it means the existing confirmations contained invalid blocks, the new confirmations are rebuilt from scratch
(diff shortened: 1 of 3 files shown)


pkg/ethblocklistener/blocklistener.go
```diff
@@ -56,6 +56,7 @@ type ConfirmationUpdateResult struct {
 	Confirmed                bool                       `json:"confirmed,omitempty"`      // when true, it means the confirmation list is complete and the transaction is confirmed
 	TargetConfirmationCount  uint64                     `json:"targetConfirmationCount"`  // the target number of confirmations for this reconcile request
 	CurrentConfirmationCount uint64                     `json:"currentConfirmationCount"` // the current number of confirmations for this reconcile request
+	Timestamp                *ethtypes.HexUint64        `json:"timestamp,omitempty"`      // the on-chain timestamp of the transaction's block - only populated in "full" chain tracking mode, since "light" mode never fetches a block
```
Chengxuan: Hi @davecrighton , the naming of the variable will need to be tightened up if it's being put at the top level.

As per my understanding, this is not when the transaction is `confirmed`, it's when the transaction is included in a block and subject to change if the chain forks etc.

`TxnBlockTimestamp` / `ReceiptBlockTimestamp`/ `ReceiptTimestamp` could be options for the name. `ReceiptTimestamp` would be my pick.


Another solution could be extending `Confirmations            []*ethrpc.MinimalBlockInfo ` to include the timestamps for each block to provide more detailed information.

Chengxuan (approved):
lgtm

PR #228, opened 2026-09-07
Backfill the in-memory canonical chain window at startup

Chengxuan:
## Summary

NOTE: this gap affects the full tracking mode + client filtering most due to the usage of `steadyStateScanCeiling` when deciding the highest block to fetch to. 

- `seedMonitoredHead` previously fetched only a single anchor block (`highestBlock-checkpointBlockGap+1`), leaving the rest of the monitored window to fill in one block at a time as the live block filter delivered new blocks - paced by real chain block production rather than polling interval.
- Right after connector startup this held back `steadyStateScanCeiling` in full chain-tracking mode, delaying client-side `getLogs` polling from reaching the chain head. It also affected `ReconcileConfirmationsForTransaction` (`validateChainCaughtUp`), which returns `FF23062` for any transaction in that gap until the window catches up - this applies to full mode regardless of `filterPollingMode` (server or client).
- `seedMonitoredHead` now fetches and reconciles the whole window up front. All but the last block are reconciled silently (no consumer notification, since they're historical, not new); the last block is still returned for the listen loop's first iteration to reconcile as before, preserving existing notification timing. On any fetch failure it falls back to the original one-block-at-a-time behavior.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] Updated `TestBlockListenerSeedMonitoredHead_BlockFound` / `_ReconcileAndDispatch` for the new return-value contract (top of window, not the anchor), with new coverage asserting the backfilled blocks land in `SnapshotMonitoredHeadChain()`


pkg/ethblocklistener/blocklistener.go
```diff
@@ -191,32 +191,57 @@ func (bl *blockListener) GetMonitoredHeadLength() int {
...
+// polling interval. On a chain with slow block times that leaves full chain-tracking mode's
+// client-side getLogs polling (steadyStateScanCeiling) holding its scan back for a long time
+// right after startup, while confirmations and re-org repair for that whole window are unavailable.
+//
+// The last (highestBlock itself) is always fetched and returned for the listen loop to reconcile on
+// its first iteration.
 func (bl *blockListener) seedMonitoredHead() *ethrpc.BlockInfoJSONRPC {
 	bl.canonicalChainLock.RLock()
 	highestBlockSet := bl.highestBlockSet
+	highestBlock := bl.highestBlock
```
peterbroadhurst, line 207: So the key change is this is still read inside the lock, but we use it outside the lock below as a threshold to preemptively catch up to blocking this function until all those blocks are fetched (or one isn't available)

pkg/ethblocklistener/blocklistener.go
```diff
@@ -191,32 +191,57 @@ func (bl *blockListener) GetMonitoredHeadLength() int {
...
+		bi := fetchBlock(blockNumber)
+		if bi == nil {
+			// Give up on the backfill and fall back to the original slow-fill behavior - the
+			// live filter will still extend the window forward from whatever it has reached
+			return nil
+		}
+		bl.reconcileCanonicalChain(bi)
+	}
+
+	bi := fetchBlock(highestBlock)
```
peterbroadhurst, line 242: Why is this outside of the loop as a separate item, vs just `<=`?

peterbroadhurst: Seems like it's because we don't call `reconcileCanonicalChain` - but why isn't that necessary just for the last block if it is for all the others?

peterbroadhurst: Ok - found this in the comments on the issue. Maybe add as a comment above this line.

> the last block is still returned for the listen loop's first iteration to reconcile as before, preserving existing notification timing

peterbroadhurst (approved):
One minor request for a comment, with approval

pkg/ethblocklistener/blocklistener.go
Chengxuan: Will add the comment.

pkg/ethblocklistener/blocklistener.go
```diff
@@ -161,7 +161,7 @@ func NewBlockListener(ctx context.Context, retry *retry.Retry, conf *BlockListen
 		isStarted:                     false,
 		startDone:                     make(chan struct{}),
 		initialBlockHeightObtained:    make(chan struct{}),
-		newHeadsTap:                   make(chan struct{}),
+		newHeadsTap:                   make(chan struct{}, 1), // buffer 1 so a tap is not dropped if the listen loop is not currently waiting
```
Chengxuan, line 164: fix intermittent test failure https://github.com/hyperledger-firefly/evmconnect/actions/runs/34322838066/job/102373183168

PR #229, opened 2026-09-08
Client side filtering updates

Chengxuan:
Follow-up to the client-side filtering work in https://github.com/hyperledger-firefly/evmconnect/pull/225 , addressing review feedback with one correctness fix and several clarity/robustness improvements.

- Fix a stall bug: es.headBlock was stored unconditionally after a poll. In full chain-tracking mode, a re-org rewind can compute a hwmBlock behind the value already stored, which would pull headBlock backward. Any listener already in individual catchup past the old value would then stall waiting for the head to catch back up to it. Added storeHeadBlockForward (CAS loop) so headBlock only ever advances, and used it in both leadGroupSteadyState and leadGroupSteadyStateGetLogs.

- Validate catchupPageSize: reject a configured value less than 1 at connector init with a new error (FF23079, MsgCatchupPageSizeInvalid) instead of letting it silently misbehave downstream.

- Renamed the "stability horizon" concept to "safe point" throughout comments, config.md, and the catchupThreshold config description, for consistency and clearer meaning (chain head minus checkpointBlockGap).

- Added tests: TestStoreHeadBlockForwardNeverRegresses, TestLeadGroupGetLogsResetsPollStateOnListenerChange, TestStreamLoopClientModeUsesGetLogsNotNewFilter, TestStreamLoopServerModeUsesNewFilter, TestLeadGroupCatchupNoDuplicateAtStableHead, plus a case in TestConnectorInit covering the new catchupPageSize validation.


internal/ethereum/event_stream.go
```diff
@@ -208,6 +208,19 @@ func (es *eventStream) catchupCeiling() (int64, bool) {
...
+// momentarily pulled the new value behind where it already was - the position must never
+// regress, or a listener in individual catchup already past it would stall until it catches
+// back up (see catchupCeiling)
+func (es *eventStream) storeHeadBlockForward(newHeadBlock int64) {
+	for {
+		current := es.headBlock.Load()
+		if newHeadBlock <= current || es.headBlock.CompareAndSwap(current, newHeadBlock) {
+			return
+		}
+	}
```
peterbroadhurst, line 224: There's a lot of words here, but they don't seem to quite match up with or explain the code.

The code seems to be "must set" which seems to be identical in function to just `es.headBlock.Store(newHeadBlock)` - because it just loops until the compare+swap turns into a `.store`.

So either:

- There's something hard to explain going on here
- There's a realization the locking granularity isn't right, and we need a mutex

internal/ethereum/event_stream.go
```diff
@@ -376,7 +390,7 @@ func (es *eventStream) leadGroupSteadyState() bool {
 				}
 
 				// Check we're not outside of the steady state window, and need to fall back to
-				// catchup mode. Catchup only polls up to the stability horizon (checkpointBlockGap
+				// catchup mode. Catchup only polls up to the safe point (checkpointBlockGap
```
peterbroadhurst: I do think "safe point" loses a bit of meaning, noting "stability horizon" isn't well described here.

I see this is a bigger find/replace, so being specific - what about "stable threshold".

peterbroadhurst (commented):
Just need a little bit more help understanding this one @Chengxuan if you don't mind giving me a little extra commentary, particularly on the difference between `while !compareAndSwapped { try(swap) }` code vs. just `store()`

internal/ethereum/event_stream.go
Chengxuan: given atomic.Int64 is used here. a mutex is not needed. The syntax above is a standard way to update value based on a conditon without overriding between concurrent thread.

1. load the old value
2. check condition is satisfied
3. attempt a write only if the current value is still the old value
4. if write was unsuccessful, it means between 1 and 3 someone else has udpate the value, go back to step `1`.

internal/ethereum/event_stream.go
Chengxuan: happy with `stable threshold`, will make the update

internal/ethereum/event_stream.go
Chengxuan: I'll add this as a code comment to explain how it works

peterbroadhurst: > check condition is satisfied

Yes agreed, that would be standard compare and swap behavior. Looking forward to the comment to explain the use here.

internal/ethereum/event_stream.go
```diff
@@ -208,6 +208,23 @@ func (es *eventStream) catchupCeiling() (int64, bool) {
...
+	// headBlock is atomic so lead-group dispatch and listener catchup can update it without
+	// taking the stream lock. CompareAndSwap(current, new) only succeeds if nobody else changed
+	// it since Load. On failure another writer won the race, so reload and retry. The loop is
+	// not a wait: it only spins on concurrent updates. We never store a lower value.
+	for {
+		current := es.headBlock.Load()
+		if newHeadBlock <= current || es.headBlock.CompareAndSwap(current, newHeadBlock) {
+			return
+		}
+	}
```
peterbroadhurst, line 224: ok - sorry I missed it. I see it there now.

Rather than the large comment block, I think this would explain things more clearly.

```suggestion
    // We do an optimistic locking update via an atomic integer (rather than using mutex)
	for {
	    // Snapshot the current value
		current := es.headBlock.Load()
		// If the current value is behind the new head, leave it unchanged
		if newHeadBlock <= current || 
		    // Otherwise attempt to update it, but if the value changed beneath us go round again 
		    es.headBlock.CompareAndSwap(current, newHeadBlock) {
			return
		}
	}
```

peterbroadhurst (approved):
Approval on there, with a suggestion for a much more compact set of comments to explain the use of optimistic concurrency update with an atomic (rather than a mutex to cover read-compare-update)

PR #231, opened 2026-09-10
Skip chain-catchup checks for zero-confirmation reconciliation

Chengxuan:
## Summary

- Moved the `targetConfirmationCount == 0` short-circuit to the very top of `ReconcileConfirmationsForTransaction`, before the chain-tracking-mode branch. It now returns a confirmed result as soon as the transaction receipt exists, the same way in both light and full mode.
- Removed `handleZeroTargetConfirmationCount` and the duplicate zero-count branch inside `buildConfirmationList`. Both are now unreachable since the top-level check handles every zero-confirmation call before execution gets that far.

## Why this is correct

A confirmation count of 0 means the caller has explicitly said "I don't need any blocks mined on top of the transaction's block, I only need it mined." Once a receipt comes back from the node, that condition is already satisfied by definition, full stop.

The old code additionally required the connector's own internal state to line up before it would say so:

- In light mode, it compared the receipt's block number against `currentChainHead`, which is this connector's cached view of the head, refreshed on its own polling/subscription cadence.
- In full mode, `validateChainCaughtUp` required the in-memory canonical chain (built by a separate block-listener subscription) to already contain a block at or past the transaction's block number.

Both of those are checks against connector-internal caches that lag the chain by design, not checks against the chain itself. The node that returned the receipt has already told us the transaction is mined. Gating a "0 confirmations needed" answer on an unrelated internal cache catching up doesn't add any real confirmation guarantee, it just introduces a race: right after startup, after a reorg, or under any polling lag, `FF23070` (`TransactionNotIncludedInChainHead`) or `FF23062` (`MsgInMemoryPartialChainNotCaughtUp`) can fire for a transaction that is unambiguously mined and requires zero further confirmation. That forces the caller (FireFly Transaction Manager) into pointless retries for a result that was already knowable from the receipt alone.

Consolidating the check into a single early return, instead of duplicating it once per chain-tracking mode, also means there is now one place that owns this behavior instead of two. Removing these checks for the `targetConfirmationCount == 0` path doesn't weaken confirmation semantics: there is nothing left to confirm once 0 confirmations are required. It only removes a spurious dependency on connector bookkeeping that was never part of the actual guarantee being requested.

## Test plan

- Removed `TestHandleZeroTargetConfirmationCount_EmptyCanonicalChain`, `TestBuildConfirmationList_NilConfirmationMap_ZeroConfirmationCountChainNotCaughtUp`, and `TestBuildConfirmationList_NewForkAfterFirstConfirmation_ZeroConfirmationCount`. They asserted behavior inside `handleZeroTargetConfirmationCount` / `buildConfirmationList` that no longer exists now that the zero-confirmation short-circuit lives one level up and those functions are never reached with a zero target.
- Kept `TestReconcileConfirmationsForTransaction_HeadBlockNumber_ZeroConfirmationCount` covering the light-mode path, with `currentChainHead` deliberately left behind the receipt block to prove the head is never consulted.
- Added `TestReconcileConfirmationsForTransaction_ZeroConfirmationCount_SkipsChainCatchup` covering the full (non-light) mode path with an empty canonical chain, to prove the in-memory partial chain is never consulted either.
- `go test ./pkg/ethblocklistener/...` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/ethblocklistener/confirmation_reconciler.go
+++ b/pkg/ethblocklistener/confirmation_reconciler.go
@@ -49,6 +49,13 @@ func (bl *blockListener) ReconcileConfirmationsForTransaction(ctx context.Contex
 			log.L(ctx).Errorf("Failed to fetch transaction receipt using tx hash %s: %v", txHash, err)
 			return nil, nil, err
 		}
+		if targetConfirmationCount == 0 {
+			return &ConfirmationUpdateResult{
+				Confirmed:                true,
+				TargetConfirmationCount:  0,
+				CurrentConfirmationCount: 0,
+			}, txReceipt, nil
+		}
 		// compare it against the chain head
 		chainHead := bl.GetHeadBlockNumber(ctx)
 		if chainHead < txReceipt.BlockNumber.Uint64() {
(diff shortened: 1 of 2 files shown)


peterbroadhurst (approved):
👍

pkg/ethblocklistener/confirmation_reconciler.go
```diff
@@ -41,6 +41,21 @@ func toBlockInfoList(ffcapiBlocks []*ethrpc.MinimalBlockInfo) (blocks []*ethrpc.
 
 func (bl *blockListener) ReconcileConfirmationsForTransaction(ctx context.Context, txHash string, existingConfirmations []*ethrpc.MinimalBlockInfo, targetConfirmationCount uint64) (*ConfirmationUpdateResult, *ethrpc.TxReceiptJSONRPC, error) {
 
+	if targetConfirmationCount == 0 {
+		// regardless of the chain tracking mode, when target confirmation count is set to 0
+		// we can immediately return a confirmed result without fetching any block information
+		txReceipt, err := bl.GetTransactionReceipt(ctx, txHash)
```
onelapahead, line 47: `txReceipt` can be `nil`, even where `err == nil`

PR #235, opened 2026-09-11
Provide flexibility on the formatting options to extend the objects

peterbroadhurst:
We have great JSON formatting features for a number of Ethereum JSON/RPC objects, but we force you to take the serialization part on the boundary of this module.

This PR proposes giving the flexibility to defer that serialization of the map so it can be extended.


EnriqueL8 (approved):
Makes sense, thanks @peterbroadhurst !

PR #236, opened 2026-09-15
fix: duplicate event dispatch when a listener has multiple filters sharing a topic0

EnriqueL8:
## Summary

Fixes duplicate event dispatch when a single listener is configured with multiple filters that share the same event signature (`topic0`) — most commonly, the same event watched across multiple contract addresses.

## Root cause

`buildAggregatedListener` indexed listeners into `listenersByTopic0` once per filter. A listener with two filters on the same event (different addresses) was inserted into that topic0's bucket twice, so `filterEnrichSort` matched and dispatched the same log twice.

Example listener config that triggers it:
```json
{
  "filters": [
    { "address": "0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", "event": { "type": "event", "name": "Transfer", "inputs": [...] } },
    { "address": "0xBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB", "event": { "type": "event", "name": "Transfer", "inputs": [...] } }
  ]
}
```

Resulting duplicate dispatch for the same event, seen twice in the logs:
```
[2026-09-11T22:34:07.101Z] DEBUG evmconnect: Detected event 000037063142/000000/000001/B=0xa6fbe9d1.../L=01a0929b-4b38-aa9d-d85b-31e897daa6b0 eventstream=019fcd50-45c6-ab9d-9dbf-74228922ec0d
[2026-09-11T22:34:07.101Z] DEBUG evmconnect: Detected event 000037063142/000000/000001/B=0xa6fbe9d1.../L=01a0929b-4b38-aa9d-d85b-31e897daa6b0 eventstream=019fcd50-45c6-ab9d-9dbf-74228922ec0d
```

## Fix

`buildAggregatedListener` now adds a listener to a given topic0 bucket at most once, regardless of how many of its filters share that topic0. The per-log match loop already checks all of a listener's filters, so this doesn't change matching behavior — only removes the duplicate bucket entry.


internal/ethereum/event_stream.go
Chengxuan: Logic-wise, it's correct. Easier to read if line 610 is moved around line 615 so it's clear what it's guarding.

Chengxuan (approved):
Great catch. Changes look good to me

internal/ethereum/event_stream.go
EnriqueL8: aah I see, yeah will move for visibility sure

PR #237, opened 2026-09-15
fix(blocklistener): Check for nil receipts on confirmations

onelapahead:
Fixes a small bug I noticed following: https://github.com/hyperledger-firefly/evmconnect/pull/231

We go and fetch a txn receipt immediately when confirmations are 0, but its possible it still hasn't been mined yet.

So `GetTransactionReceipt` returns `nil`, and we pass that along marking it as `confirmed: true` when it hasn't been yet.


Chengxuan (approved):
LGTM

pkg/ethblocklistener/confirmation_reconciler.go
```diff
@@ -49,6 +49,10 @@ func (bl *blockListener) ReconcileConfirmationsForTransaction(ctx context.Contex
 			log.L(ctx).Errorf("Failed to fetch transaction receipt using tx hash %s: %v", txHash, err)
 			return nil, nil, err
 		}
+		if txReceipt == nil {
+			log.L(ctx).Debugf("Transaction receipt not yet available for tx hash %s", txHash)
+			return nil, nil, i18n.NewError(ctx, msgs.MsgTransactionNotFound, txHash)
+		}
```
Chengxuan, line 59: Thinking about this again. This should be a `confirmed: false` situation, not an error.

Chengxuan (changes requested):

pkg/ethblocklistener/confirmation_reconciler.go
onelapahead: I'm not sure thats right:
```go
&ConfirmationUpdateResult{
  Confirmed:                false,
  TargetConfirmationCount:  0,
  CurrentConfirmationCount: 0,
}
```

the fact `target == current` is smelly to me. And later on we error like this for a similar cases:
```go
		if chainHead < txReceipt.BlockNumber.Uint64() {
			// the transaction is not yet included in the chain head
			return nil, nil, i18n.NewError(ctx, msgs.MsgTransactionNotIncludedInChainHead, txHash, chainHead, txReceipt.BlockNumber.String())
		}
		
		// other branches of code
		
	if txBlockInfo == nil {
		log.L(ctx).Debugf("Transaction %s not found in any block", txHash)
		return nil, nil, i18n.NewError(ctx, msgs.MsgTransactionNotFound, txHash)
	}
```

only once we have a tx receipt, do we call:

```go
	confirmationUpdateResult, err := bl.buildConfirmationList(ctx, blockInfoExistingConfirmations, txBlockInfo, targetConfirmationCount)
	if confirmationUpdateResult != nil {
		confirmationUpdateResult.TargetConfirmationCount = targetConfirmationCount
		confirmationUpdateResult.CurrentConfirmationCount = uint64(len(confirmationUpdateResult.Confirmations)) - 1
		confirmationUpdateResult.TxnBlockTimestamp = &txBlockInfo.Timestamp
		// NOTE: This function does not do the full receipt decoding, for which there is a complex function for.
		// The "Receipt" object is left empty (but the JSON/RPC receipt is return to the caller for enrichment)
	}
```

which can set `Confirmed: false` based on the counts.

onelapahead: Its likely either is right/OK - just pointing out this feels slightly inconsistent. I'm analyzing tests/callers to see if there would be any material difference.

onelapahead: Okay confirmed this likely won't make any difference in all the callers, but the inconsistency with `txBlockInfo` is  the only smell.

Which would you now prefer @Chengxuan ?

Chengxuan: Those are great points.


The key reason errors are kept separate is to inform consumers of unexpected situations. e.g., failing to gather all required info, failing to invoke the RPC endpoint; the in-memory partial chain not yet synced with the chain state

For the situation where a transaction has no receipt. It doesn't fall into the same category.



--------------

 ```
&ConfirmationUpdateResult{
  Confirmed:                false,
  TargetConfirmationCount:  0,
  CurrentConfirmationCount: 0,
}
```

I can see why that feels weird `Confirmed = txReceipt != nil && TargetConfirmationCount === CurrentConfirmationCount`. But txReceipt is not in the object.

A few options for this special case:

1. Accept the ugliness of it. Consumers that already checks txReceipt !== nil before looking into the confirmation result should not notice this. But we should update `confirmed: false` to be correct for consumers that use `confirmationUpdateResult` directly.
2. return `nil` for  `confirmationUpdateResult` when txnReceipt is `nil`, this could break existing consumers that only care about `confirmationUpdateResult`
3. Move txReceipt into `confirmationUpdateResult`, and maybe also omit `CurrentConfirmationCount` when `TargetConfirmationCount` is 0.


@onelapahead I think both `1` and `3` are acceptable, and happy for `3` to become a follow-on PR.

onelapahead: I think I will just do 1) as that is acceptable for consumers from what I have seen.

Chengxuan (approved):
LGTM Thanks @onelapahead

PR #240, opened 2026-09-21
Remove use of eth_newBlockFilter with client-side filtering

peterbroadhurst:
We added an option for client-side filtering in EVMConnect in #225 and #229, which removed use of `eth_newFilter` and `eth_getFilterChanges` for detection of EVM logs/events.

However, we were still using `eth_newBlockFilter` / `ethgetFilterChanges` for detection of blocks in `full` mode.

This PR updates the block listener nodes too as follows:

## Block polling modes

`events.filterPollingMode` (`server` | `client`) now governs the block listener as well as event filters. Combined with `chainTrackingMode` (`full` | `light`) this gives four modes, each implemented as its own poller behind a shared listen loop:

| `chainTrackingMode` | `filterPollingMode` | New blocks discovered by | Canonical chain |
|---|---|---|---|
| `full` | `server` | `eth_newBlockFilter` + `eth_getFilterChanges`; each hash resolved with `eth_getBlockByHash` | Built from the filter's hashes. Gaps (e.g. after a lost filter) and re-orgs are repaired by the chain rebuild. |
| `full` | `client` | `eth_getBlockByNumber("latest")` each poll; no node-side state | Built from the head block alone. A head that does not fit on the tail triggers the same chain rebuild, which fills by number from the last valid block. |
| `light` | `server` | `eth_blockNumber` | None. Only the head number is dispatched. |
| `light` | `client` | `eth_blockNumber` | None. Only the head number is dispatched. |

Notes:

- Light mode no longer creates a block filter in `server` mode (previously it created one and discarded the hashes).
- In full mode the monitored head window is seeded before the loop starts and nothing is dispatched for it; the first event consumers see carries `GapPotential=true` in every mode.
- Steady-state cost is one RPC call per poll in every mode. The `newHeads` WebSocket subscription, where configured, only wakes the poll early; it never supplies block data.


pkg/ethblocklistener/blocklistener.go
peterbroadhurst, line 331: I started with a much more patch-like change to the `listenerLoop`, but the 4 different paths diverge so much in their function that it was pretty impossible to read.

I hope the approach of splitting out the separate discrete `poll()` functions is easier to read & validate.

pkg/ethblocklistener/blocklistener.go
```diff
@@ -340,161 +337,38 @@ func (bl *blockListener) listenLoop() {
...
+// waitNextPoll backs off after a failure, otherwise sleeps for the polling interval (or until we're
+// shoulder tapped by the newHeads listener). Returns false when the listener is stopping.
+func (bl *blockListener) waitNextPoll(failCount int, firstIteration bool) bool {
+	if failCount > 0 {
+		return !bl.retry.DoFailureDelay(bl.ctx, failCount)
+	}
+	if firstIteration {
+		return true
+	}
+	return bl.waitNextIteration()
```
peterbroadhurst, line 361: Note we still only use the `newHeads` as a shoulder-tap. So in client+full mode, we get a `newHeads`, discard the block hashes we're notified with (which might or might not be the latest) and do a full `eth_getBlockByNumber("latest", false)`

pkg/ethblocklistener/blocklistener_pollers.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// newBlockPoller picks the poller for the configured modes, and does the one-off work before the loop
+// starts: seeding the canonical chain, and marking the listener started where no filter has to be
+// established first (consumers can only be added once started).
+func (bl *blockListener) newBlockPoller() blockPoller {
+	if bl.ChainTrackingMode == ffcapi.ChainTrackingModeLight {
+		bl.markStarted()
+		return &headNumberPoller{bl: bl}
+	}
```
peterbroadhurst, line 48: Light mode is always just the number (client/server filtering doesn't mean anything in this mode, as we're not allowed to get the blocks via either mechanism in light mode).

pkg/ethblocklistener/blocklistener_pollers.go
```diff
@@ -0,0 +1,214 @@
...
+	bl.seedMonitoredHead()
+	if bl.FilterPollingMode == FilterPollingModeClient {
+		bl.markStarted()
+		return &latestBlockPoller{canonicalChainPoller{bl: bl, gapPotential: true}}
+	}
+	return &blockFilterPoller{canonicalChainPoller: canonicalChainPoller{bl: bl, gapPotential: true}}
```
peterbroadhurst: We don't `markStarted()` until `eth_newBlockFilter` has returned on the block filter path

internal/ethereum/config.go
```diff
@@ -58,17 +59,17 @@ const (
...
-	// FilterPollingModeClient uses stateless eth_getLogs range queries, with the connector tracking
-	// its own in-memory poll position - avoiding node-side filter state entirely
-	FilterPollingModeClient filterPollingMode = "client"
+	// FilterPollingModeServer uses node-side filters, established with eth_newBlockFilter / eth_newFilter
+	// and polled with eth_getFilterChanges, so the node tracks what is new since the last poll
+	FilterPollingModeServer = ethblocklistener.FilterPollingModeServer
+	// FilterPollingModeClient polls the latest block, and uses stateless eth_getLogs range queries, with
+	// the connector tracking its own in-memory poll positions - avoiding node-side filter state entirely
+	FilterPollingModeClient = ethblocklistener.FilterPollingModeClient
 )
```
Chengxuan, line 73: Assuming duplication here is for documentation purposes only.

pkg/ethblocklistener/blocklistener_pollers.go
```diff
@@ -0,0 +1,214 @@
...
+	bl.dispatchToConsumers(bl.snapshotConsumers(), update)
+	return nil
+}
+
+// canonicalChainPoller is the part shared by the full chain tracking pollers: reconciling the blocks a
+// poll discovered into the canonical chain, and notifying consumers from the lowest changed position.
+type canonicalChainPoller struct {
+	bl           *blockListener
+	gapPotential bool // true until the first successful poll, and again while a lost block filter is re-established
+}
```
Chengxuan: Bad naming, this is not a poller and doesn't implement `poll()`.

pkg/ethblocklistener/blocklistener_pollers.go
```diff
@@ -0,0 +1,214 @@
...
+// by one block fills the cache for the hashes after it.
+func (bl *blockListener) resolveBlockHashes(blockHashes []ethtypes.HexBytes0xPrefix) iter.Seq[*ethrpc.BlockInfoJSONRPC] {
+	return func(yield func(*ethrpc.BlockInfoJSONRPC) bool) {
+		for _, h := range blockHashes {
+			if bi := bl.resolveBlockHash(h); bi != nil && !yield(bi) {
+				return
+			}
+		}
+	}
+}
```
Chengxuan, line 160: took me a while to understand the usage of iter.Seq to defer block lookup here. looks good

Chengxuan (approved):
Changes look good to me. One minor comment on the naming of canonicalChainPoller

pkg/ethblocklistener/blocklistener_pollers.go
peterbroadhurst: 👍 shall see if I can simplify

pkg/ethblocklistener/blocklistener_pollers.go
peterbroadhurst: Made it `canonicalChainPollerBase`

internal/ethereum/config.go
peterbroadhurst: Yeah - it also removes a lot of long references throughout the code in this package

pkg/ethblocklistener/blocklistener_pollers.go
peterbroadhurst: Yeah - it's the first use of this go 1.23 interface in the package, and the first time I've used it.
Go going a bit Java with iterators 😄

